PR #29, opened 2022-06-21
Optimized images

TheAnonyUnknown:


gokadzev:
just for 35kb it doesn't worth to make worse assets quality.

PR #209, opened 2023-04-15
update vietnamese locali

giangphan11:
add-Traditional-vietnamese


pubspec.yaml
```diff
@@ -31,7 +31,7 @@ dependencies:
   hive: ^2.2.3
   hive_flutter: ^1.1.0
   http: ^0.13.5
-  intl: ^0.17.0
```
gokadzev, line 34: This shouldn't be changed, it causes conflict.

pubspec.lock
```diff
@@ -92,10 +92,10 @@ packages:
     dependency: transitive
     description:
       name: characters
-      sha256: e6a326c8af69605aec75ed6c187d06b349707a27fbff8222ca9cc2cff167975c
```
gokadzev, line 95: since we only add language, neither pubspec.lock should be changed.

giangphan11: Because musify depends on flutter_localizations from sdk which depends on intl 0.18.0, intl 0.18.0 is required.
So, because musify depends on intl ^0.17.0, version solving failed.

gokadzev: > Because musify depends on flutter_localizations from sdk which depends on intl 0.18.0, intl 0.18.0 is required. So, because musify depends on intl ^0.17.0, version solving failed.

yes, you need to revert change (intl: ^0.17.0 -> intl: ^0.18.0), intl version should be ^0.17.0 otherwise it causes errors.

gokadzev: wait, does it shows 0.18.0 for you? it shows 0.17.0 for me

lib/main.dart
```diff
@@ -41,6 +41,7 @@ final appLanguages = <String, String>{
   'Traditional Chinese Taiwan': 'zh_TW',
   'Turkish': 'tr',
   'Ukrainian': 'uk',
+  'Tiếng Việt': 'vi',
```
gokadzev, line 44: I think it would be better if we named it Traditional Vietnamise

pubspec.lock
giangphan11: i downgraded intl version from 0.18.0 to 0.17.0

gokadzev: Please also undo changes in pubspec.lock

gokadzev:
It looks good now, thank you for contribution :)

PR #253, opened 2023-06-19
Update app_ru.arb

S0nter:
Translated "Recently Played"


gokadzev:
**Update**: New translations are added

PR #266, opened 2023-07-20
Added Greek translations

CrisBalGreece:
Please add Greek translations when free..
Thank you for your hard work there..


gokadzev (approved):
Thanks!

PR #296, opened 2023-10-01
CustomPlaylists load again!

captaincrunch42:
Modified so Custom Playlists load again (and are not stuck) and custom playlists are removable!  The problem was that the conversion of "yitd" to String made it so that it's content is 'null' instead of null, which rendered multiple checks useless. My guess is that the change was done because the playlistRemoveFunction requiers a non nullable string. Therefore a new (working) remove function was also written.

### DIFF
--- a/lib/screens/user_added_playlists_page.dart
+++ b/lib/screens/user_added_playlists_page.dart
@@ -171,12 +171,13 @@ class _UserPlaylistsPageState extends State<UserPlaylistsPage> {
                   ),
                   itemBuilder: (BuildContext context, index) {
                     final playlist = playlists[index];
-                    final ytid = playlist['ytid'].toString();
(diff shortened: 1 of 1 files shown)


captaincrunch42:
Maybe a confirmation alert before removing a CustomUserPlaylist would be good so that it is not lost forever by a simple missclick.

gokadzev:
> Modified so Custom Playlists load again (and are not stuck) and custom playlists are removable! The problem was that the conversion of "yitd" to String made it so that it's content is 'null' instead of null, which rendered multiple checks useless. My guess is that the change was done because the playlistRemoveFunction requiers a non nullable string. Therefore a new (working) remove function was also written.

Thank you! I have no idea how I missed this.

PR #299, opened 2023-10-04
Bring package back to AndroidManifest.xml

nna-dev:
Bring package back to AndroidManifest.xml to bypass "No application found for TargetPlatform.android_arm64" error when build the app on M1 chipset


gokadzev:
Strange, it was removed automatically while upgrading AGP and worked on both my personal pc and workflow.  But as it causes problem in some cases, i will revert it everywhere.

PR #424, opened 2024-09-29
Update app_de.arb

Yoshlix:
Updated German translation


gokadzev:
@Yoshlix I just noticed the punctuation marks at the end of the sentences. If the original file (app_en.arb) doesn't have a punctuation mark at the end (there are several exceptions where I have left them, and I'll take a look later), it shouldn't be included in the translation either, as the punctuation is already fulfilled in the code. Could you please make this correction in the translation?

Yoshlix:
I corrected it.

PR #446, opened 2024-11-20
Update app_pt.arb

GeovaneSec:


gokadzev:
@GeovaneSec Hi, Thanks for this PR. I had to add one more localization that I'm going to push in several minutes, will you change it too or I can merge this commit already?

GeovaneSec:
Yes, I can update the upcoming commits with that localization as well.

gokadzev:
> Yes, I can update the upcoming commits with that localization as well.

Perfect, no need to close tho, just make changes in the same branch and it'll appear here too

GeovaneSec:
I have made a new pull, with only one commit.

PR #456, opened 2024-12-05
Make user interface more intuitive

Heinrich-XIAO:
My first time using Musify, I thought it was a bug that it wouldn't open the full song play menu. Turns out, I was supposed to swipe up. I've never seen this before and the only reason I found out I was supposed to swipe up was from reading the code.


gokadzev:
Yeah but with this PR nothing changes, it still remains intuitive + swipe left and right changing music, so opening full player page on tap will be annoying while doing that. Maybe I'll think about something in V9.

Issue #462, opened 2024-12-07
Playing only parts of a song 


Heinrich-XIAO:
### Is there an existing issue for this?

- [X] I have searched the existing issues and confirmed that this feature is not already requested.

### Category of the Feature

Unable to categorize

### Credits

Me

### Pain

I have to manually stop the song where I want the song to end. I sometimes only like some of the song but then I have to start and stop it myself. 

### Suggested Solution

It would be great if there was a "song profile" so I can play only some parts of the song. A song profile should be a profile to control how a song gets played ie the start and end times.

Heinrich-XIAO:
A song profile should be able to be used for other things as well like audio balancing, and other aspects of playing songs.

gokadzev:
That's not really something I'd like to work on. If it's about extra seconds coming from videos, SponsorBlock does its job.

Heinrich-XIAO:
Its not about ads or anything its just that there are parts of songs that I like more than others. Its about playing different songs in different ways.

Heinrich-XIAO:
Some songs are inherently louder than others and I want to have it play quieter so I have a consistent volume throughout a playlist.

Heinrich-XIAO:
Also, I could work on this myself.

gokadzev:
> Some songs are inherently louder than others and I want to have it play quieter so I have a consistent volume throughout a playlist.

Either I don't understand what you mean and I thought I did or you're suggesting several features here. I don't see any connection between this and what you said above

Heinrich-XIAO:
You may not be understanding it. I'm saying you would be able to change how songs are played in volume, start time, end time and more.

Heinrich-XIAO:
I already have it done, but not with the start and end times parts because I started with volume because I thought it would've been easier. I will still do the start and end times if this feature is deemed useful.

PR #463, opened 2024-12-08
Song profile

Heinrich-XIAO:
#462

### DIFF
--- a/lib/localization/app_en.arb
+++ b/lib/localization/app_en.arb
@@ -90,5 +90,6 @@
   "undo": "Undo",
   "useSquigglySlider": "Squiggly slider",
   "userPlaylists": "User playlists",
+  "volumeSongProfile": "Volume",
   "youtubePlaylistLinkOrId": "YouTube playlist link or ID"
 }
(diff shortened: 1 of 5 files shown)


Heinrich-XIAO:
This is pretty much done. Tell me if I should change anything.

gokadzev:
Individual volume control for each song? doesn't make any sense and haven't seen anywhere such thing. sorry, I'm still pessimistic towards to this feature.

Heinrich-XIAO:
I know it's not anywhere else (to my knowledge) but it is useful to some people (at least me).

Heinrich-XIAO:
It let's you have a playlist go without having to change volume each time, this is especially useful if you listen to music sleeping as you are more sensitive to small changes in volume.

gokadzev:
> It let's you have a playlist go without having to change volume each time, this is especially useful if you listen to music sleeping as you are more sensitive to small changes in volume.

Well, this PR is broken, and a separate page for it wouldn't be a good idea for the volume feature. This means we need to integrate it into the Now Playing page, which is quite complex as that page has been unresponsive on different screens for a long time. Even now, I'm not sure it looks perfect on every screen size.

Heinrich-XIAO:
Well, I'm going to add more options to that page.

Issue #472, opened 2024-12-23
[Bug] Shuffle


Heinrich-XIAO:
### Acknowledgements

- [X] I have searched the existing issues and confirmed that this is a new issue, not a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to the latest version. You can find the latest version [here](https://github.com/gokadzev/Musify/releases).
- [X] I will fill out all the sections in this issue report.

### Current Behavior

Shuffle does not continue picking random songs when playing last song. 

### Expected Behavior

It should keep choosing random songs in the playlist. 

### Steps To Reproduce

1. Get playlist of more than one song.
2. Play until playing the last song.
3. Wait till the song finishes or put the player to the end
4. It just stops unlike every other time. 

### Musify Version

Latest

### Device And Operating System

Huawei

### Android Version You Use

Android 15

gokadzev:
> ### Acknowledgements
> * [x]  I have searched the existing issues and confirmed that this is a new issue, not a duplicate or related to another open issue.
> * [x]  I have written a short but informative title.
> * [x]  I have updated the app to the latest version. You can find the latest version [here](https://github.com/gokadzev/Musify/releases).
> * [x]  I will fill out all the sections in this issue report.
> 
> ### Current Behavior
> Shuffle does not continue picking random songs when playing last song.
> 
> ### Expected Behavior
> It should keep choosing random songs in the playlist.
> 
> ### Steps To Reproduce
> 1. Get playlist of more than one song.
> 2. Play until playing the last song.
> 3. Wait till the song finishes or put the player to the end
> 4. It just stops unlike every other time.
> 
> ### Musify Version
> Latest
> 
> ### Device And Operating System
> Huawei
> 
> ### Android Version You Use
> Android 15
> 
> ### Additional Information
> _No response_

That's not shuffle but repeat function for playlist that isn't added tho

Heinrich-XIAO:
No, shuffle actually does repeat, just stops at the last one, it's theoretically possible for it to just play the first few forever, but whenever it reaches the last song it halts.

gokadzev:
> No, shuffle actually does repeat, just stops at the last one, it's theoretically possible for it to just play the first few forever, but whenever it reaches the last song it halts.

I think you might have misunderstood what shuffle feature means.

Heinrich-XIAO:
It is exactly what it does except for the last song. It plays random songs forever, but here it plays forever until it gets to the last song by shear chance.

[...]

PR #473, opened 2024-12-23
Fix shuffle bug

Heinrich-XIAO:
#472


Heinrich-XIAO:
Tested

Heinrich-XIAO:
now that  https://github.com/gokadzev/Musify/commit/da5cc2726497afacfcd0aed49e7cdd8a91e34d0b i have to do this again.

PR #476, opened 2024-12-25
fix: shuffle bug in #472

Heinrich-XIAO:


gokadzev:
this does nothing but plays the first song again, so PR is invalid.

Heinrich-XIAO:
@gokadzev did you not read the comment? it plays a random song, which it originally didn't when it played the last song.

gokadzev:
> @gokadzev did you not read the comment? it plays a random song, which it originally didn't when it played the last song.

I did and both shuffle and random song picker features work, I checked myself.

PR #517, opened 2025-02-26
feat: add Chinese translation

jbwfu:
I'm sorry, but I don't understand Dart, so I'm not sure if my modifications contain errors. Since Chinese has both simplified and traditional characters, I can't use "zh" or "cn" for naming, and I'm also unsure if my naming is appropriate.


gokadzev:
@jbwfu I merged it already but it shows this:

```bash
Error: Arb file for a fallback, zh, does not exist, even though 
the following locale(s) exist: [zh_Hans, zh_Hant]. 
When locales specify a script code or country code, a 
base locale (without the script code or country code) should 
exist as the fallback. Please create a {fileName}_zh.arb 
file.
```

we need to have app_zh.arb too, should we rename either of these or there's more generalized variant of Chinese considered as a standard?

gokadzev:
I have an idea, no need to do anything from your side. thanks :)

jbwfu:
> we need to have app_zh.arb too, should we rename either of these or there's more generalized variant of Chinese considered as a standard?

The more widely used variant is zh-Hans.

PR #524, opened 2025-03-04
Fix call resume behavior

Ruchan10:
Currently the app resumes the music after a call ends even if the music was paused before call. 
a simple change in _initialize function of audio_service.dart fixes the issue. Added wasPlayingBeforeInterruption variable to remember the music state.


gokadzev:
Hi, there were lot's of unrelated changes in this pull request. feel free to open an issue about it so I'll see what's the best way to solve this problem

PR #532, opened 2025-03-08
Fix: Music resumes after call, only if it was playing before

Ruchan10:
## Problem
- When a call comes in, the music pauses automatically, which is fine.
- However, even if the music was **paused before the call**, it **resumes playing** after the call ends.
- This means the user has to manually pause it again, which is unexpected behavior.

## Expected Behavior
- If the music was **already paused before the call**, it should **remain paused** after the call ends.
- If the music **was playing before the call**, it should **resume playback** as expected.

## Fix
- Store the playback state before the call.
- Only resume playback if it was playing before the call.

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -53,6 +53,7 @@ class MusifyAudioHandler extends BaseAudioHandler {
 
   Timer? _sleepTimer;
   bool sleepTimerExpired = false;
+  late bool wasPlayingBeforeCall = false;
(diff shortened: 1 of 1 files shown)


gokadzev:
It should be fixed with the commit: https://github.com/gokadzev/Musify/commit/e1492024a80189f9986e7af3e6c92b14f5e702cd

It seems like it handles all cases pretty well if you just configure the session type.

PR #543, opened 2025-03-13
Turkish translations patch 2

BinaryHarbinger:
* Corrected some translations.


gokadzev:
* symbols in translations should be matched with English ones (so no . or ! if it's not in English version)
* it would be better if you recreate branch with just 1 commit with meaningful name with all these changes.

PR #577, opened 2025-05-26
fix(ui): adjust spacing in mobile layout by commenting out SizedBox

cychowai:
Some UI issues related to the device bar covers the app area, so using a safearea to resolve the issue

Screenshot 2025-05-26 at 9 19 31 PM
Screenshot 2025-05-26 at 9 22 58 PM

After applying the safearea, there is a hardcode sized container with a height that causes the widget overflow, so removing it.

Screenshot 2025-05-26 at 9 19 11 PM
Screenshot 2025-05-26 at 9 19 05 PM
Screenshot 2025-05-26 at 9 28 18 PM

After applying these changes, I have tested with android devices of different sizes, and it should be fine!

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -142,6 +142,7 @@ class _MusifyState extends State<Musify> {
         const SystemUiOverlayStyle(
           statusBarColor: Colors.transparent,
           systemNavigationBarColor: Colors.transparent,
+          systemStatusBarContrastEnforced: true,
         ),
       );
     });
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -187,28 +187,30 @@ class _MusifyState extends State<Musify> {
 
   @override
   Widget build(BuildContext context) {
-    return DynamicColorBuilder(
-      builder: (lightColorScheme, darkColorScheme) {
-        final colorScheme = getAppColorScheme(
-          lightColorScheme,
-          darkColorScheme,
-        );
+    return SafeArea(
+      child: DynamicColorBuilder(
+        builder: (lightColorScheme, darkColorScheme) {
+          final colorScheme = getAppColorScheme(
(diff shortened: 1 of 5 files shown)


lib/screens/now_playing_page.dart
```diff
@@ -177,7 +177,7 @@ class _MobileLayout extends StatelessWidget {
             iconSize: adjustedMiniIconSize,
             isLargeScreen: isLargeScreen,
           ),
-          const SizedBox(height: 35),
+          // const SizedBox(height: 35),
```
gokadzev, line 180: it's exactly on the navigation bar then if we don't add any space here, so we need it but maybe less than 35

lib/main.dart
```diff
@@ -187,28 +187,30 @@ class _MusifyState extends State<Musify> {
 
   @override
   Widget build(BuildContext context) {
-    return DynamicColorBuilder(
-      builder: (lightColorScheme, darkColorScheme) {
-        final colorScheme = getAppColorScheme(
-          lightColorScheme,
-          darkColorScheme,
-        );
+    return SafeArea(
```
gokadzev, line 190: if we use SystemUiOverlayStyle, we need to add systemStatusBarContrastEnforced: true, in SystemUiOverlayStyle too

gokadzev:
Now another issue appears, system navigation bar becomes black instead of app theme background color (as it's right now without this PR)

gokadzev:
also I think you accidentally picked d70d9a446754ec5f96c6afee9419740dc77bc9a5 commit, if that's so it needs to be reverted

gokadzev:
@cychowai That's it, reverting d70d9a446754ec5f96c6afee9419740dc77bc9a5 and pulling latest changes should be enough

PR #579, opened 2025-05-26
Fix/safe area picked

cychowai:
Sorry for merging the incorrect master to sync with the latest codes,
Now the commits are picked


cychowai:
Checked

gokadzev:
If not, we’ll have to keep this PR open, because at the moment we’re either losing the transparent status and navigation bars for SafeArea, or vice versa.

Of course until I find a solution.

gokadzev:
I'll directly make changes in your branch

gokadzev:
> Ok, thanks

Thank you so much!

PR #584, opened 2025-05-27
Fix/offline n bottom nav

cychowai:
restart the app after the change of offline mode
fixing the custom submit button does not using the submit text passing into the component, always show "cancel"
fix the navigation on the bottom does not update after the change of offline mode
fix the selected index is not correct for the navigation after changing the offline mode


cychowai:
@gokadzev Could you check again? Please take the whole PR for testing, instead of cherry pick, thank you

cychowai:
Ok let me share my testing result with you later

PR #588, opened 2025-05-28
refactor: update confirmation dialog to use submitBtnText for consist…

cychowai:
…ency

feat: add restart_app dependency for app restart functionality


cychowai:
@gokadzev I found out that the restart app package cannot update the route in the navigation manager. I tried many ways, but it is hard to update the route without a big revamp of the navigation of the whole project, so I decided to just keep my branch, and will not fix the offline issues at the moment.

PR #611, opened 2025-07-22
Estonian language translation

Wqttzicue:
Im unsure if I found correct localization file.
Im native Estonian speaker. I may make changes later because i could theoretically wrongly understand context, but it should be fine.
I dont know how to actually add the language to app so itll be actually used, but I hope you can add it where needed yourself


gokadzev:
Hi, thanks! I'll take care of the rest.

PR #622, opened 2025-08-14
fixed song list update issue

Arjunkvl:
The initial seach worked fine but after that the 'song' Listview wasnt updating it was because the list was similar and flutter coudn't recognise any change. I added a 'key' field in Listview.builder which solved the issue.

i hope you will have a look.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .buildlog/
 .history
 .svn/
+.fvm/
 
 # IntelliJ related
 *.iml
--- a/lib/screens/search_page.dart
+++ b/lib/screens/search_page.dart
@@ -96,7 +96,6 @@ class _SearchPageState extends State<SearchPage> {
     if (_fetchingSongs.value) {
       _fetchingSongs.value = false;
     }
-
     setState(() {});
   }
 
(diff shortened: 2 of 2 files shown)


gokadzev:
Ios, macos, linux platform folders should be removed. Also last time I checked it was not only list visibility issue but also API issue, it was throwing redirection loop detected or something like that. Does it perform 3-5 searches without any problem?

4rjxn:
I have removed the folders. For me only the first search worked then the song part dosent change. But albums and playlist worked fine.

4rjxn:
Do I also need to remove the .gitignore file.

I don't know much about PR this is my first one  😄.

dust63:
> Ios, macos, linux platform folders should be removed. Also last time I checked it was not only list visibility issue but also API issue, it was throwing redirection loop detected or something like that. Does it perform 3-5 searches without any problem?

I'm new user, and on first usage I have the bug too.

Step 1: make the search with one title
Step 2: make a new search with another one title
The result still display first search result.
Step 3: make a search with empty string. The result now is empty.
Step 4: make a search again and the result now is ok.

4rjxn:
Exactly this was the bug I had.
and I solved it with key `key:` argument

gokadzev:
> Exactly this was the bug I had. and I solved it with key `key:` argument

It was not completely UI state update problem as I said, so I'm working on different approach that will be pushed in several minutes along with new release build

PR #624, opened 2025-09-28
make the github release action honor the flutter version from pubspec.yaml

IzzySoft:
v9.6.3 failed RB, because your release action wasn't pinned to the flutter version specified in `pubspec.yaml`, and thus used the "currently latest stable available". This PR adjust the workflow, based on the fdroid one, so future releases will be RB again without manual intervention (hopefully :wink:)

For 9.6.3, we've manually corrected it in our build recipe, so it is RB again.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,10 +44,13 @@ jobs:
             ${{ runner.os }}-gradle-
 
       # Setup the Flutter environment.
-      - name: Set up Flutter
-        uses: subosito/flutter-action@v2
+      - run: |
+          pip install -U yq
+          echo "FLUTTER_VERSION=$(yq -r .environment.flutter pubspec.yaml | sed 's/\^//g')" >> $GITHUB_ENV
+      - uses: subosito/flutter-action@v2
         with:
           channel: 'stable'
+          flutter-version: "${{ env.FLUTTER_VERSION }}"
 
       #
       - run: |


gokadzev:
Thanks!

gokadzev:
Wait, if it compares the F-Droid build with the GitHub release build, does that mean I had a separate F-Droid build in GitHub releases for no reason?

IzzySoft:
> Wait, if it compares the F-Droid build with the GitHub release build, does that mean I had a separate F-Droid build in GitHub releases for no reason?

You have 2 different workflows there. At IzzyOnDroid, we use the APK produced by the fdroid one. Not sure what the differences are between the two flavors – but I guess you didn't establish them "for no reason" :wink:

gokadzev:
> > Wait, if it compares the F-Droid build with the GitHub release build, does that mean I had a separate F-Droid build in GitHub releases for no reason?
> 
> You have 2 different workflows there. At IzzyOnDroid, we use the APK produced by the fdroid one. Not sure what the differences are between the two flavors – but I guess you didn't establish them "for no reason" 😉

[Musify.apk](https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify.apk) - Universal build with built-in updater.
[Musify-arm64-v8a.apk](https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-arm64-v8a.apk) - smaller apk for only arm64 devices and it comes with built-in updater as well
[Musify-fdroid.apk](https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-fdroid.apk) - F-droid build with no built-in updater.


And this is the fdroid build workflow where I already had this change added: https://github.com/gokadzev/Musify/blob/master/.github/workflows/fdroid.yml

IzzySoft:
Oh… now, that's funny. So I adjusted the wrong action? Here's what we use:

```yaml
  - tag: 9.6.3
    apks:
      - apk_pattern: Musify-arm64-v8a\.apk
        apk_url: https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-arm64-v8a.apk
        build:
          - sed -r '/signingConfigs.getByName/d ; /create\("release"\) \{/,/        \}/d' -i android/app/build.gradle.kts
          - FLUTTER_VERSION="$(sed -rn 's/\s*\bflutter:\s\^?([0-9.]+)\s*$/\1/p' pubspec.yaml)"
          - FLUTTER_VERSION="3.35.4"
          - git clone -b $FLUTTER_VERSION https://github.com/flutter/flutter
          - export PUB_CACHE=$(pwd)/.pub-cache
          - flutter/bin/flutter config --no-analytics
          - flutter/bin/flutter pub get
          - flutter/bin/flutter build apk --release --split-per-abi --target-platform android-arm64 --flavor github
          - mv build/app/outputs/apk/github/release/app-github-arm64-v8a-release-unsigned.apk /outputs/unsigned.apk
        build_cpus:
        build_home_dir: /home/runner
        build_repo_dir: /home/runner/work/Musify/Musify
...
```

And as you can see, I had to override the flutter version to make it RB – which means, your build did **not** use the flutter version specified in `pubspec.yaml`. Can you please check?

> [Musify-arm64-v8a.apk](https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-arm64-v8a.apk) - smaller apk for only arm64 devices and it comes with built-in updater as well

Oopsie? `REQUEST_INSTALL_PACKAGES` was removed there back in March. If it still has a built-in updater, we might have an issue there: unless explicitly opt-in and explaining where it comes from and what implications it has (namely, bypassing extra screening in our repo here), that would be against our inclusion policy.

gokadzev:
> Oh… now, that's funny. So I adjusted the wrong action? Here's what we use:
> 
> ```yaml
>   - tag: 9.6.3
>     apks:
>       - apk_pattern: Musify-arm64-v8a\.apk
>         apk_url: https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-arm64-v8a.apk
>         build:
>           - sed -r '/signingConfigs.getByName/d ; /create\("release"\) \{/,/        \}/d' -i android/app/build.gradle.kts
>           - FLUTTER_VERSION="$(sed -rn 's/\s*\bflutter:\s\^?([0-9.]+)\s*$/\1/p' pubspec.yaml)"
>           - FLUTTER_VERSION="3.35.4"
>           - git clone -b $FLUTTER_VERSION https://github.com/flutter/flutter
>           - export PUB_CACHE=$(pwd)/.pub-cache
>           - flutter/bin/flutter config --no-analytics
>           - flutter/bin/flutter pub get
>           - flutter/bin/flutter build apk --release --split-per-abi --target-platform android-arm64 --flavor github
>           - mv build/app/outputs/apk/github/release/app-github-arm64-v8a-release-unsigned.apk /outputs/unsigned.apk
>         build_cpus:
>         build_home_dir: /home/runner
>         build_repo_dir: /home/runner/work/Musify/Musify
> ...
> ```
> 
> And as you can see, I had to override the flutter version to make it RB – which means, your build did **not** use the flutter version specified in `pubspec.yaml`. Can you please check?
> 
> > [Musify-arm64-v8a.apk](https://github.com/gokadzev/Musify/releases/download/9.6.3/Musify-arm64-v8a.apk) - smaller apk for only arm64 devices and it comes with built-in updater as well
> 
> Oopsie? `REQUEST_INSTALL_PACKAGES` was removed there back in March. If it still has a built-in updater, we might have an issue there: unless explicitly opt-in and explaining where it comes from and what implications it has (namely, bypassing extra screening in our repo here), that would be against our inclusion policy.

As I explained, I specifically created two flavors to respect F-Droid policy, so --flavor github is for non-fdroid users who need built-in updater and --flavor fdroid is for F-Droid users who don't need it.

About REQUEST_INSTALL_PACKAGES, it's not even used in the non-fdroid builds (As I just use dialog and then if user clicks on download, it's redirected in browser and starts downloading the release). If the code above is from the F-droid workflow (not in my repo but in the actual F-Droid one) it means it was configured incorrectly this whole time. On my side everything looks fine, even in the workflow: https://github.com/gokadzev/Musify/blob/master/.github/workflows/fdroid.yml#L69

IzzySoft:
Then we have an issue here – unless the updater is configured as I outlined above. Can you please say if it meets those criteria? Otherwise, we'd have to switch to the bigger fdroid APK, which would only allow us to keep one release here (as it's almost 30 MB, which is our per-app limit) – and could soon even go beyond the size limit, raising the next issue.

Also, this still leaves the question why your build used a different flutter version than the one defined in `pubspec.yaml`, which is a problem for RB. Shouldn't it have [picked the flutter version from `pubspec.yaml`](https://github.com/gokadzev/Musify/blob/master/.github/workflows/release.yml#L46)?

gokadzev:
> Then we have an issue here – unless the updater is configured as I outlined above. Can you please say if it meets those criteria? Otherwise, we'd have to switch to the bigger fdroid APK, which would only allow us to keep one release here (as it's almost 30 MB, which is our per-app limit) – and could soon even go beyond the size limit, raising the next issue.
> 

I'm not sure what you're asking in the first part, if it's about REQUEST_INSTALL_PACKAGES, none of the builds have it. the only difference is github API fetch function and giving a dialog that says update is available with the download button.

> Also, this still leaves the question why your build used a different flutter version than the one defined in `pubspec.yaml`, which is a problem for RB. Shouldn't it have [picked the flutter version from `pubspec.yaml`](https://github.com/gokadzev/Musify/blob/master/.github/workflows/release.yml#L46)?

Well, we just found out that F-Droid uses my regular github arm-64 apk instead of separately created fdroid flavor which I've been releasing this whole time. I can add split apk command in fdroid workflow so It releases arm64 instead of universal builds for F-Droid.

gokadzev:
release.yaml is for my GitHub users’ builds (with universal and arm64 APKs that include an update dialog).
fdroid.yaml is for F-Droid builds, which have no update dialog.

IzzySoft:
Got that, thanks – but that still leaves the question whether the IzzyOnDroid inclusion criteria are met with the Github build. You seem to avoid answering that, so let me ask directly for a yes/no:

1. is the update check opt-in, does it clarify where the updates are taken from, and does it point out the implications?
2. if the answer to 1. is "no", can we have an arm64-only build for the fdroid variant as well, so we can switch to that at IzzyOnDroid?
3. alternatively to 2., could you make the update check optional, with outlining source and implications in the dialog?

Just trying to find common ground here. We've got a responsibility towards those using our repository, so we need to ensure :wink:

gokadzev:
> Got that, thanks – but that still leaves the question whether the IzzyOnDroid inclusion criteria are met with the Github build. You seem to avoid answering that, so let me ask directly for a yes/no:
> 
> 1. is the update check opt-in, does it clarify where the updates are taken from, and does it point out the implications?
> 2. if the answer to 1. is "no", can we have an arm64-only build for the fdroid variant as well, so we can switch to that at IzzyOnDroid?
> 3. alternatively to 2., could you make the update check optional, with outlining source and implications in the dialog?
> 
> Just trying to find common ground here. We've got a responsibility towards those using our repository, so we need to ensure 😉

I didn’t avoid any questions, though. Anyway:

1. It was never intended for F-Droid, so it’s clear that if a user installs that build, it still comes from the same source. As I already mentioned, neither Musify.apk nor Musify-arm64.apk were meant for F-Droid. The dialog looks like this:

image

2. That’s what I’ve been suggesting. I always knew F-Droid was using Musify-Fdroid.apk; I never thought it was relying on the normal builds (which only differ in the API check and the update dialog, once again). So yes, I could do that.

3. I think the second approach would be better (changing the F-Droid universal APK to the F-Droid ARM64 APK). It wouldn’t cause any installation conflicts, and realistically, who would install it from F-Droid and then update it using GitHub?

gokadzev:
@IzzySoft 

here 9.6.4 version with arm64 apk: https://github.com/gokadzev/Musify/releases/tag/9.6.4

and here are the workflow changes I just made:
https://github.com/gokadzev/Musify/commit/b74a48b438ffa1cfe7b8f14ce445e7a3f7d9a782

IzzySoft:
> I didn’t avoid any questions, though.

Apologies: a notification got lost (in the spam folder), so the next one jumped right to the next comment, which made me miss one.

> I'm not sure what you're asking in the first part

The paragraph of the [IzzyOnDroid App Inclusion Policy](https://gitlab.com/IzzyOnDroid/repo/-/wikis/Inclusion-Policy), almost literally shared with F-Droid's:

> [the app] must not download additional executable binary files (e.g. addons, auto-updates, etc.) without explicit user consent. Consent means it needs to be opt-in (it must not be harder to decline than to accept or presented in a way users are likely to press accept without reading) and structured in a way that clearly explains to users that they’re choosing to bypass the checks performed in this repo if they activate it.

> That’s what I’ve been suggesting. I always knew F-Droid was using Musify-Fdroid.apk; I never thought it was relying on the normal builds (which only differ in the API check and the update dialog, once again). So yes, I could do that.

I don't know which APK F-Droid uses, or which flavor it builds – but I can check quickly…

```yaml
Binaries: https://github.com/gokadzev/Musify/releases/download/%v/Musify-fdroid.apk
```

Yupp, the fdroid flavor, with the `.fdroid` suffix. What I'm talking about here however is [the listing at IzzyOnDroid](https://apt.izzysoft.de/packages/com.gokadzev.musify), existing since 2022-06-12, and also established as reproducible build.

OK, so if we use option 2, that would mean the packageName would change, which would make it a different app, so we had to "relocate". Can be done of course, if you prefer that. The text on the updater prompt would not be enough to meet the inclusion criteria (especially not if the update check is enabled by default), thanks for the screenshot.

> realistically, who would install it from F-Droid and then update it using GitHub?

Well, maybe those waiting for an update stuck in the build queue there – especially if it's an important update? F-Droid takes at least 2-5 days until an update you make shows up, sometimes longer when the timing was unlucky or there are build server issues. IzzyOnDroid usually ships updates within 24h. "Build server issues" here would only affect reproducible builds – which at IzzyOnDroid run on a separate track, not slowing down any releases to be shipped.

> I think the second approach would be better (changing the F-Droid universal APK to the F-Droid ARM64 APK). It wouldn’t cause any installation conflicts

What installation conflicts? Musify is established RB at F-Droid, so they ship the APK you've built and signed. IzzyOnDroid always ships the APKs built and signed by their resp. developers. So all have the same signature.

Using option 2 would mean that everyone who'd installed Musify from IzzyOnDroid, would no longer receive updates from us, but had to uninstall and reinstall to switch to the fdroid build – which is why we'd prefer option 3. We won't press for it, however, as both variants would comply with our inclusion criteria – so the decision is yours. Just let us know what it should be; should you need a little time to think it over, that's of course OK (but as we are already in a "non-compliant state", we shouldn't wait too long).

Thanks a lot!

> here 9.6.4 version with arm64 apk: [9.6.4 (release)](https://github.com/gokadzev/Musify/releases/tag/9.6.4)

Please remember to have the universal one for F-Droid (`Musify-fdroid.apk`) built and attached as well, or they can't pick up the release as their RB will fail. You could of course also build the other ABIs and attach them, then ask F-Droid to switch to those for future releases; I guess they'd even welcome that.

And yupp, that APK has the `.fdroid` suffix in the packageName, so it would require everyone to "switch over". If that's what you want, we can do that.

gokadzev:
> > realistically, who would install it from F-Droid and then update it using GitHub?
> 
> Well, maybe those waiting for an update stuck in the build queue there – especially if it's an important update? F-Droid takes at least 2-5 days until an update you make shows up, sometimes longer when the timing was unlucky or there are build server issues. IzzyOnDroid usually ships updates within 24h. "Build server issues" here would only affect reproducible builds – which at IzzyOnDroid run on a separate track, not slowing down any releases to be shipped.
> 
> > I think the second approach would be better (changing the F-Droid universal APK to the F-Droid ARM64 APK). It wouldn’t cause any installation conflicts
> 
> What installation conflicts? Musify is established RB at F-Droid, so they ship the APK you've built and signed. IzzyOnDroid always ships the APKs built and signed by their resp. developers. So all have the same signature.
> 
> Using option 2 would mean that everyone who'd installed Musify from IzzyOnDroid, would no longer receive updates from us, but had to uninstall and reinstall to switch to the fdroid build – which is why we'd prefer option 3. We won't press for it, however, as both variants would comply with our inclusion criteria – so the decision is yours. Just let us know what it should be; should you need a little time to think it over, that's of course OK (but as we are already in a "non-compliant state", we shouldn't wait too long).
> 
> Please remember to have the universal one for F-Droid (`Musify-fdroid.apk`) built and attached as well, or they can't pick up the release as their RB will fail. You could of course also build the other ABIs and attach them, then ask F-Droid to switch to those for future releases; I guess they'd even welcome that.
> 
> And yupp, that APK has the `.fdroid` suffix in the packageName, so it would require everyone to "switch over". If that's what you want, we can do that.

Oh wait you're talking about IzzyOnDroid, I thought we stopped releasing updates there since we moved on F-Droid?

gokadzev:
Now I’m confused. Will the changes I made break F-Droid builds? I thought the problem was with F-Droid, I had no idea it was also released on IzzyOnDroid.


If it does, I can revert those changes and we can try adding:  -t lib/main_fdroid.dart in the build command.

I use  `flutter build apk --release --split-per-abi --target-platform="android-arm64" --flavor fdroid -t lib/main_fdroid.dart` command to build fdroid app and it adds isFdroid variable with true value so the app hides everything related to the updater (dialog and check updates tile in the settings page)


edit: Nevermind, I think IzzyOnDroid also checks reproducibility so it won't work

IzzySoft:
> Oh wait you're talking about IzzyOnDroid, I thought we stopped releasing updates there since we moved on F-Droid?

Part of the team [left there](https://gitlab.com/fdroid/admin/-/issues/447) almost 2 years ago, and is since focusing on IzzyOnDroid. We re-established some of the apps that were "moved over" and still met our inclusion criteria. Apologies if I caused confusion here! I've mentioned that fact [above](#issuecomment-3344180387), but probably not clear enough…

> Will the changes I made break F-Droid builds?

Only if you do no longer provide the universal build of their flavor.

> I think IzzyOnDroid also checks reproducibility

Indeed we do; that is why I opened this PR in the first place, because our RB failed with version differences. My bad that I didn't check before _which_ exact flavor we were using; I just found that one workflow takes the flutter version from `pubspec.yaml` while the other didn't, so I thought it must be that…

We can adjust our build recipe as well, to follow any changes you'd need to make here. You can find our recipe [in this comment](https://github.com/gokadzev/Musify/pull/624#issuecomment-3345590570) – just let me know what we should adjust, and we can do that. For example:

>  I can revert those changes and we can try adding: `-t lib/main_fdroid.dart` in the build command.

Easy on our end as well. I just think it would preferable if we had _not_ to change the packageName here.

gokadzev:
> > Oh wait you're talking about IzzyOnDroid, I thought we stopped releasing updates there since we moved on F-Droid?
> 
> Part of the team [left there](https://gitlab.com/fdroid/admin/-/issues/447) almost 2 years ago, and is since focusing on IzzyOnDroid. We re-established some of the apps that were "moved over" and still met our inclusion criteria. Apologies if I caused confusion here! I've mentioned that fact [above](#issuecomment-3344180387), but probably not clear enough…
> 
> > Will the changes I made break F-Droid builds?
> 
> Only if you do no longer provide the universal build of their flavor.
> 
> > I think IzzyOnDroid also checks reproducibility
> 
> Indeed we do; that is why I opened this PR in the first place, because our RB failed with version differences. My bad that I didn't check before _which_ exact flavor we were using; I just found that one workflow takes the flutter version from `pubspec.yaml` while the other didn't, so I thought it must be that…
> 
> We can adjust our build recipe as well, to follow any changes you'd need to make here. You can find our recipe [in this comment](https://github.com/gokadzev/Musify/pull/624#issuecomment-3345590570) – just let me know what we should adjust, and we can do that. For example:
> 
> > I can revert those changes and we can try adding: `-t lib/main_fdroid.dart` in the build command.
> 
> Easy on our end as well. I just think it would preferable if we had _not_ to change the packageName here.

Yeah, it sounds the best solution is dialog that asks user if they want to receive updates and if user clicks no, they won't get any update related dialogs. Should I hide 'check updates' button in the settings too?

IzzySoft:
> Yeah, it sounds the best solution is dialog that asks user if they want to receive updates and if user clicks no, they won't get any update related dialogs.

:+1: And include where updates come from, and that they'd bypass any additional screening e.g. at IzzyOnDroid.

> Should I hide 'check updates' button in the settings too?

No need, as long as that explanation is next to it. That way we could simply stick with the very same github flavor you use here anyway, would make it easier for you then, too, right?

gokadzev:
> > Yeah, it sounds the best solution is dialog that asks user if they want to receive updates and if user clicks no, they won't get any update related dialogs.
> 
> 👍 And include where updates come from, and that they'd bypass any additional screening e.g. at IzzyOnDroid.
> 
> > Should I hide 'check updates' button in the settings too?
> 
> No need, as long as that explanation is next to it. That way we could simply stick with the very same github flavor you use here anyway, would make it easier for you then, too, right?

Done, it also mentions that it's being downloaded from GitHub releases

Screenshot_20250929-215333_com android systemui
Screenshot_20250929-215815_com gokadzev musify debug

IzzySoft:
Thanks! Though it's missing the pointer of the implications for those having installed it via e.g. IzzyOnDroid.

So for IzzyOnDroid, we stick with the arm64 APK of the github flavor as before then?

PR #656, opened 2025-11-15
Share song option

ryanfox1985:
Added a new option to share a song with other apps.

Screenshot_20251115-192350
Screenshot_20251115-192359

### DIFF
--- a/lib/API/musify.dart
+++ b/lib/API/musify.dart
@@ -1121,6 +1121,9 @@ Future<bool> makeSongOffline(dynamic song, {bool fromPlaylist = false}) async {
         return false;
       }
 
+      song['audioBitrate'] = '${audioManifest.bitrate.kiloBitsPerSecond} kbps';
+      song['audioCodec'] = audioManifest.codec.mimeType;
+
       final stream = _yt.videos.streamsClient.get(audioManifest);
       final fileStream = audioFile.openWrite();
       await stream.pipe(fileStream);


gokadzev:
Hi, so it's basically a download feature, which I refused to add just because it sounds pointless + it will have lower volume. Using YouTube-to-MP3 converters makes much more sense than using this feature to save music as a file.

ryanfox1985:
Yep I agree that could be redundant with downloading, I'm trying to learn play music and this share feature saves / make easier to me to share the song with other apps like https://moises.ai or DAW

ryanfox1985:
about the quality, I was wondering if there is a better quality than `m4a` and then the downlading is already using the best quality, isn't? 

- https://github.com/ryanfox1985/Musify/blob/play_playlist/lib/API/musify.dart#L964
- https://github.com/ryanfox1985/Musify/blob/play_playlist/lib/API/musify.dart#L1114

gokadzev:
> about the quality, I was wondering if there is a better quality than `m4a` and then the downlading is already using the best quality, isn't?
> 
> * https://github.com/ryanfox1985/Musify/blob/play_playlist/lib/API/musify.dart#L964
> * https://github.com/ryanfox1985/Musify/blob/play_playlist/lib/API/musify.dart#L1114

Just try downloading any songs (since you can modify the code and test on your device), then compare how they (exact same songs) sound in any other local music player app versus how they sound in Musify, so you'll know what I mean.

ryanfox1985:
ok, I checked the codec and bitrate and this YT lib with the public YT API is fetching a poor quality... like `mp3 / 128kbps` then online sites could give you `mp3 / 320 kbps`.

online sites what they probably do are:
- use YT private API.
- Download video and then extract audio from it.

ryanfox1985:
what would do you think in a new option in `settings => audio quality` like `ultra high (high data transfer and processing)` and then download the video and extract the audio?

gokadzev:
> what would do you think in a new option in `settings => audio quality` like `ultra high (high data transfer and processing)` and then download the video and extract the audio?

FFmpeg used to be a complete headache for me, and as far as I know, there’s no actively maintained library for it right now. If someone else contributes it, it might get merged, but another issue that comes up is the app size, which would increase significantly.

ryanfox1985:
mmm yeah, this option of ffmpeg seems an overkill... I guess the problem is using free services like this YT lib bypassing paid service...

The only alternative to download and parse video in mobile side is to pay YT or use a proxy (but I guess these sites wants to make business with ads too)...

PR #657, opened 2025-11-16
Feature play playlist.

ryanfox1985:
https://photos.app.goo.gl/dyQ3GNs4KBme5E7q6


gokadzev:
The idea itself isn’t too bad (even if it’s somewhat redundant, since the user can just click the first song and it will play automatically). But it means we'd need a new design for the playlist page even more urgently than we do now, because it's already overloaded with buttons.

ryanfox1985:
Mmm I didn't realised about that, seems not very intuitive...

PR #703, opened 2026-02-09
Improve Home UI + Add Settings + Glassmorphism Mode

AnujFlutter:
What’s changed

Updated Home screen UI (Scroller view improvements)

Added new Settings section

Added Glass Mode feature (iOS-style glass effect)

Glass Mode

When enabled:

Mini player widget shows as a glass card

Bottom navigation bar switches to glass style


gokadzev:
@AnujFlutter Hi, I noticed that this PR includes changes to quite a lot of files, including some icon files. Could you check that all of these changes are intended and related to the PR? also if possible, please attach screenshots what the app looks like after those changes are made

AnujFlutter:
https://github.com/user-attachments/assets/859faca1-3663-4c8c-96dc-6c82a2fc18c4

Screenshot_20260209-190703_Musify

gokadzev:
@AnujFlutter Hmm, I like the glass effect, but I don’t really like how it looks with Material 3. Maybe we could introduce a new theme that works better with the glass style.

But first, I'd need to see a cleaner PR without all the extra file changes, so I can understand what the glass-related code actually looks like. Right now, your PR shows 195 files changed, which doesn't seem logical for this kind of changes.

AnujFlutter:
Hi @gokadzev,
Thanks for your feedback! I understand your concern regarding Material 3 and the glass effect.
I will clean up the PR and remove all unrelated file changes so that only the glass-effect-related modifications remain. Once done, I’ll update the pull request for clearer review.
Also, I’m open to introducing a new theme for the glass style if you feel that would be a better approach with Material 3.
Let me know if you have any specific expectations for the cleanup or theme design.
Thanks again for your review!

gokadzev:
> Hi @gokadzev, Thanks for your feedback! I understand your concern regarding Material 3 and the glass effect. I will clean up the PR and remove all unrelated file changes so that only the glass-effect-related modifications remain. Once done, I’ll update the pull request for clearer review. Also, I’m open to introducing a new theme for the glass style if you feel that would be a better approach with Material 3. Let me know if you have any specific expectations for the cleanup or theme design. Thanks again for your review!

The best approach in this case would be to open PR with just glass effect applied to bottom navigation bar and mini player (as shown in screenshots) with toggle in settings, and then when I'm free enough to work on it, I'll merge your changes + make additional commits after for full optional app transition to the new design (I'm not sure if we manage to do it without changing codebase much, so that's why I'm asking to leave only PR-related changes for now and then we will see :) )

**Update comment:**

Challenges when adding a glassmorphism theme:
- Ensuring it works well in dark mode
- Keeping the codebase maintainable
- Preventing any impact on existing theme modes

PR #749, opened 2026-03-08
feat: add offline search functionality

elias001011:
## Description
This PR adds a much-needed offline search feature for downloaded and liked songs. 

Previously, if the user was offline, the Search tab would disappear or become unusable. Now, users can easily search through their locally saved music library without needing an internet connection.

## Changes made
- **`bottom_navigation_page.dart`**: Removed the condition that hid the Search tab when `isOfflineMode` was active. The tab is now always visible. Removed the fallback that forced navigation to the Home tab when going offline.
- **`search_page.dart`**: Added `_intelligentLocalSearch` to filter through `userOfflineSongs` and `userLikedSongsList`. It splits the search query into words and matches them against track titles, artists, and albums regardless of word order.
- **UI Updates**: Added visual sections for "Offline Songs" and "Liked Songs" using the existing localization keys (`context.l10n!.offlineSongs` and `context.l10n!.likedSongs`), so translations work out of the box via Crowdin without requiring new `.arb` entries.

## How to test
1. Enable Offline Mode in the app settings (or turn off Wi-Fi/Data).
2. Go to the Search tab (it will now be visible).
3. Type the name, artist, or album of a song you have previously downloaded or liked.
4. The results will populate instantly under the "Offline Songs" and "Liked Songs" headers.

### DIFF
--- a/lib/screens/bottom_navigation_page.dart
+++ b/lib/screens/bottom_navigation_page.dart
@@ -143,22 +143,13 @@ class _BottomNavigationPageState extends State<BottomNavigationPage> {
         route: '/home',
         shellIndex: 0,
       ),
-    ];
-
-    // Only add search tab in online mode
-    if (!isOfflineMode) {
-      items.add(
-        _NavigationItem(
-          icon: FluentIcons.search_24_regular,
-          selectedIcon: FluentIcons.search_24_filled,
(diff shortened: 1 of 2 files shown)


gokadzev:
The approach we discussed has been added in v9.9.0. Thank you so much for your contribution!

PR #758, opened 2026-03-08
feat(user_songs): add shuffle play button for library categories

elias001011:
## Description
This PR introduces a **Shuffle play** button to the header section of the User Songs pages (Liked songs, Offline songs, and Recently played).

## Changes made
- Added a new `_buildShufflePlayButton` widget alongside the existing Play and Clear buttons.
- The function securely creates a copy of the current songs list, applies Dart's native `.shuffle()`, and replaces the current queue with it.
- Follows the app's current UI design by utilizing `IconButton.filledTonal` and `FluentIcons.arrow_shuffle_24_filled`.

This is a small but very useful feature, especially for users with large offline/liked libraries who want a quick way to listen to their tracks in random order. Thank you!

### DIFF
--- a/lib/screens/user_songs_page.dart
+++ b/lib/screens/user_songs_page.dart
@@ -189,6 +189,7 @@ class _UserSongsPageState extends State<UserSongsPage> {
           runSpacing: 8,
           children: [
             if (songsLength > 0) _buildPlayButton(primaryColor, title),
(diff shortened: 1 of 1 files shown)


gokadzev:
Ah, I had it added for other playlists but not for those. Thanks!

Issue #744, opened 2026-03-04
[Feature] "Create a playlist" button when adding songs via bigplayer


erickmufath:
### Is there an existing issue for this?

- [x] I have searched the existing issues and confirmed that this feature is not already requested.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this feature.

### Category of the Feature

UX Improvements

### Credits

Me

### Pain

I often use the "Link Sharing" feature, but it's a bit of a hassle because it doesn't allow me to create playlists directly. I have to go to the Library to create a playlist and back again to bigplayer

### Suggested Solution

![Image](https://github.com/user-attachments/assets/15452d5e-29c3-491f-890b-7c2bd9da53f2)
![Image](https://github.com/user-attachments/assets/7a826200-0bb2-4c59-8379-25c62054c8ae)

Add buttons, icons, or text to create a playlist there.

erickmufath:
@elias001011 If you want, you can check the closed issues, regardless of closed as completed or closed as not planned, there are some ideas from me that cannot be fully explained well (I use translator btw). Maybe you can contribute and help @gokadzev.

elias001011:
@erickmufath I can take a look when I have some free time. This project is very useful to me, and I’m trying to help @gokadzev with issues that seem interesting from a user experience perspective. Thank you for your feedback.

erickmufath:
> [@erickmufath](https://github.com/erickmufath) I can take a look when I have some free time. This project is very useful to me, and I’m trying to help [@gokadzev](https://github.com/gokadzev) with issues that seem interesting from a user experience perspective. Thank you for your feedback.

I would really appreciate it. You can filter by my username or search for issues containing my username for a wider reach.

PR #759, opened 2026-03-08
Feature/add playlist shortcut

elias001011:
### What does this PR do?
This PR introduces a quick shortcut to create a new playlist and automatically add a specific song to it, directly from the "Add to Playlist" dialog inside `SongBar`.

### Why is this needed?
Previously, if a user wanted to add a song to a new custom playlist that didn't exist yet, they had to:
1. Close the song menu.
2. Go to the Library page.
3. Create the playlist.
4. Go back to the song.
5. Open the menu again and add it.

Now, users can just tap "Add Playlist" inside the song menu, and the dialog will default to the "Custom" tab, create the playlist, and insert the song in one go.

### Code Refactoring details:
To avoid code duplication, I extracted the `_showAddPlaylistDialog` function from `library_page.dart` into a new shared utility file (`lib/utilities/playlist_dialogs.dart`). Now both the `LibraryPage` and `SongBar` use this same widget, keeping the codebase clean.

- **Library Page:** Continues to work exactly as before (opens defaulting to YouTube mode).
- **Song Menu:** Opens the same dialog but defaults to Custom Mode (since it receives the `songToAdd` parameter) and handles the insertion right after creation.


elias001011:
I was inspired by Issue #744 to add this feature. I thought it was super simple and I think it would make life easier for new users. Maybe my approach wasn't the best, but I tried to implement something that wouldn't duplicate code.

lib/utilities/playlist_dialogs.dart
```diff
@@ -0,0 +1,318 @@
...
+              FilledButton.icon(
+                onPressed: () async {
+                  if (isYouTubeMode && id.isNotEmpty) {
+                    final result = await addUserPlaylist(id, context);
+                    if (context.mounted) showToast(context, result);
+                    
+                    if (songToAdd != null && context.mounted) {
+                       // The newly added youtube playlist will be the last one
+                       if (userPlaylists.value.isNotEmpty) {
+                           final newPlaylist = userPlaylists.value.last;
```
gokadzev: we declare this variable but never use it, so it should be either used or removed.

gokadzev (changes requested):

elias001011:
@gokadzev Addressed the review feedback in the latest commits.

Changes made:
- Renamed `showAddPlaylistDialog` to `showCreatePlaylistDialog` to make its purpose clearer.
- Updated the related call sites in `library_page.dart` and `song_bar.dart`.
- Removed the unused variable in the YouTube playlist flow.
- Fixed the accidental unrelated changes introduced while updating `song_bar.dart`.

The feature behavior remains the same:
- Library page opens the dialog normally.
- Song menu can create a new custom playlist and add the selected song right after creation.

PR #760, opened 2026-03-08
Feature/playlist image

### LINKED ISSUE #740: [Feature] Make an image from a song as a playlist image
### Is there an existing issue for this?

- [x] I have searched the existing issues and confirmed that this feature is not already requested.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this feature.

### Category of the Feature

UX Improvements

### Credits

Me

### Pain

The playlist that was created was left without an image, too lazy to add it manually either from local or link.

### Suggested Solution

Image

Add options like "set image as playlist image" at the three dots.

Algorithm to prevent bugs:
- Show options only when open a playlist and song are in a playlist
- Show options only when the song has an image/thumbnail and meet the requirements
- The playlist referred to is only the playlist that is opened
- If the song used as the playlist image is deleted or moved, delete the playlist image as well and leave it blank

### Useful Resources/References

_No response_

elias001011:
### Description
This PR introduces the ability to use a specific song's artwork as the cover image for a custom playlist. It was inspired by and addresses the feature requested in issue #740. 

### Changes made
- Added a new **"Set as playlist image"** option in the song's popup menu (`song_bar.dart`), which appears when viewing a custom playlist.
- Implemented logic in `playlists_manager.dart` to save the selected image url/path to the playlist data (supporting both standalone custom playlists and those inside folders).
- Added logic to automatically clear the playlist image if the song providing the artwork is removed from that playlist.
- Added a **"Remove Image"** button in the `EditPlaylistDialog` so users can easily clear the cover if they change their minds.
- Handled image selection for both offline (`artworkPath`, `highResImage`) and online songs.
- Added EN and PT localization strings for the new UI elements.

### ⚠️ Note / Known Limitation
**OBS:** Please note that for some specific songs, the image simply fails to load. When the user tries to set one of these songs as the playlist cover, the app sets the image as empty because the image link is unavailable/broken. I couldn't find a definitive workaround for this specific edge case. I'm leaving this note here in case you have any ideas on how to handle this fallback, or if you want to review this behavior before merging!

Closes #740


gokadzev (changes requested):

gokadzev:
@elias001011 It's not a problem. If you want, I can make those changes after merging if it's more convenient for you.

elias001011:
@gokadzev Addressed in the latest commit.

Changes made in `song_bar.dart`:
- Simplified the image selection logic as suggested.
- Replaced the previous `.then().catchError()` flow with an async/await flow.
- Added an HTTP reachability check for remote image URLs before calling `setPlaylistImageFromSong`.
- Kept local paths working as before.
- Added proper user feedback:
  - success toast when the playlist image is updated,
  - `noImageFound` when the song image URL is invalid/unreachable,
  - `setAsPlaylistImageError` for unexpected failures.

This also resolves the edge case I mentioned before: for songs whose image URL is broken/unavailable, the app now detects that the image is invalid instead of setting an empty playlist image.

gokadzev:
@elias001011 I just noticed that this PR not only adds playlist image support but also adds an "Add as a playlist image" option to the song bar, which I'm pessimistic about, to be honest.

elias001011:
@gokadzev Thanks for the feedback.

I just want to clarify the expected scope before I remove anything, because Issue #740 seems to specifically request adding a “set as playlist image” action in the song options menu.

That is why I implemented the SongBar option together with the playlist image support. If you would prefer a different UX, could you clarify what behavior you want to keep from this PR and how you expect the playlist image to be changed instead?

I can adjust the PR, but I want to make sure I align it with your intended direction first.

gokadzev:
> @gokadzev Thanks for the feedback.
> 
> I just want to clarify the expected scope before I remove anything, because Issue #740 seems to specifically request adding a “set as playlist image” action in the song options menu.
> 
> That is why I implemented the SongBar option together with the playlist image support. If you would prefer a different UX, could you clarify what behavior you want to keep from this PR and how you expect the playlist image to be changed instead?
> 
> I can adjust the PR, but I want to make sure I align it with your intended direction first.

Yeah, but it leads to more issues than benefits, such as an overloaded dropdown and unstable image URLs. Even if we have checks, they can still be low quality or inaccessible, and users may start complaining about it later, etc. That's why I think it's the best if we don't add it, especially when users still can set artworks by using URLs from anywhere

gokadzev:
It's better if we focus on existing feature, finding and fixing bugs in them and UI/UX improvements rather than adding more features I guess.

elias001011:
> It's better if we focus on existing feature, finding and fixing bugs in them and UI/UX improvements rather than adding more features I guess.

Thanks for clarifying.

Just to make sure I understood correctly: do you want me to remove the SongBar button entirely and keep only the underlying playlist image support?

My concern is that, without the button, this PR may become redundant from my point of view, since the main requested behavior in Issue #740 seemed to be setting the playlist image directly from a song.

If that is indeed the direction you want, I can either:
- remove only the SongBar UI/action and keep the remaining playlist image-related changes, or
- close the PR if you feel the remaining scope is no longer useful.

gokadzev:
> > It's better if we focus on existing feature, finding and fixing bugs in them and UI/UX improvements rather than adding more features I guess.
> 
> Thanks for clarifying.
> 
> Just to make sure I understood correctly: do you want me to remove the SongBar button entirely and keep only the underlying playlist image support?
> 
> My concern is that, without the button, this PR may become redundant from my point of view, since the main requested behavior in Issue #740 seemed to be setting the playlist image directly from a song.
> 
> If that is indeed the direction you want, I can either:
> 
> * remove only the SongBar UI/action and keep the remaining playlist image-related changes, or
> * close the PR if you feel the remaining scope is no longer useful.

You're right, I was confused initially and that's why I asked to remove some parts. Sorry but closing this PR would be better. Thank you for your contributions tho, in case of bug-fixes or the most problematic todos (https://github.com/gokadzev/Musify/discussions/504) feel free to open pull requests again :)

PR #770, opened 2026-03-10
fix(ui): aliigment setSleepTimer text

elias001011:
In some languages, the ‘Set Sleep Timer’ text was overflowing or going outside the container, so I fixed that. This was happening in several languages where the translation was longer.

**Before**

**After**

### DIFF
--- a/lib/widgets/now_playing/bottom_actions_row.dart
+++ b/lib/widgets/now_playing/bottom_actions_row.dart
@@ -14,7 +14,6 @@
  *     You should have received a copy of the GNU General Public License
  *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
  *
- *
  *     For more information about Musify, including how to contribute,
  *     please visit: https://github.com/gokadzev/Musify
  */
@@ -302,11 +301,13 @@ void _showSleepTimerDialog(BuildContext context) {
               children: [
                 Icon(FluentIcons.timer_24_regular, color: colorScheme.primary),
                 const SizedBox(width: 12),
-                Text(
-                  context.l10n!.setSleepTimer,
-                  style: TextStyle(
-                    color: colorScheme.onSurface,
-                    fontWeight: FontWeight.w600,
+                Expanded(
+                  child: Text(
+                    context.l10n!.setSleepTimer,
+                    style: TextStyle(
+                      color: colorScheme.onSurface,
+                      fontWeight: FontWeight.w600,
+                    ),
                   ),
                 ),
               ],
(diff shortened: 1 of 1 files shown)


gokadzev:
@elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.

elias001011:
> @elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.

Yep. 

> @elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.

I agree. So would just shortening the translation solve the problem? Anyway, I tried to make a correction in this PR that covered the current situation. I did another translation PR for pt-br that solves some long translations into Portuguese, but the "SetSleepTimer" is still long there. I can adjust that.

gokadzev:
> > @elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.
> 
> I agree. So would just shortening the translation solve the problem? Anyway, I tried to make a correction in this PR that covered the current situation. I did another translation PR for pt-br that solves some long translations into Portuguese, but the "SetSleepTimer" is still long there. I can adjust that.

About translations, could you please use Crowdin instead? (https://crowdin.com/project/musify)

It’s more convenient - everything is synced and doesn’t conflict with anything. Also, if you want, I can give you access to one of the social media accounts (Discord) I created for Musify but didn’t really use, so you can tell me what you’re going to work on. That way, in similar cases, you won’t waste time submitting PRs that might not be merged.

By the way, thanks for improving the app!

elias001011:
> > > @elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.
> > 
> > 
> > I agree. So would just shortening the translation solve the problem? Anyway, I tried to make a correction in this PR that covered the current situation. I did another translation PR for pt-br that solves some long translations into Portuguese, but the "SetSleepTimer" is still long there. I can adjust that.
> 
> About translations, could you please use Crowdin instead? (https://crowdin.com/project/musify)
> 
> It’s more convenient - everything is synced and doesn’t conflict with anything. Also, if you want, I can give you access to one of the social media accounts (Discord) I created for Musify but didn’t really use, so you can tell me what you’re going to work on. That way, in similar cases, you won’t waste time submitting PRs that might not be merged.
> 
> By the way, thanks for improving the app!

I would be happy for that, actually. I use the app every day and I just want to help improve it. Thank you

gokadzev:
> > > > @elias001011 I think the problem is that even in the English translations we use "Set a sleep timer" instead of just "Sleep timer", which sounds more appropriate for a dialog UI.
> > > 
> > > 
> > > I agree. So would just shortening the translation solve the problem? Anyway, I tried to make a correction in this PR that covered the current situation. I did another translation PR for pt-br that solves some long translations into Portuguese, but the "SetSleepTimer" is still long there. I can adjust that.
> > 
> > 
> > About translations, could you please use Crowdin instead? (https://crowdin.com/project/musify)
> > It’s more convenient - everything is synced and doesn’t conflict with anything. Also, if you want, I can give you access to one of the social media accounts (Discord) I created for Musify but didn’t really use, so you can tell me what you’re going to work on. That way, in similar cases, you won’t waste time submitting PRs that might not be merged.
> > By the way, thanks for improving the app!
> 
> I would be happy for that, actually. I use the app every day and I just want to help improve it. Thank you

my username is: gov2025

PR #781, opened 2026-03-13
feat(player): hydrate standalone playback queue from YouTube Music

Shrawan13-glitch:
## Summary
- add a YouTube Music watch-next queue hydrator for standalone playback
- integrate queue generation into the audio handler so non-playlist playback grows into a contextual queue
- improve rapid skip handling so the latest requested target wins without briefly playing intermediate tracks
- seed a fresh queue again near queue end for non-playlist playback
- remove the automatic song picker setting and related logic to avoid conflicts with the new queue flow
- update the debug workflow to use the current artifact action version

## Testing
- flutter analyze
- manual testing of standalone playback queue generation
- manual testing of rapid skip behavior
- manual testing of queue growth near queue end

## Testing notes
- Play next works as expected
- Add to queue works as expected
- The earlier rapid-skip issue where intermediate songs briefly played has been fixed in my testing

### DIFF
--- a/.github/workflows/debug.yml
+++ b/.github/workflows/debug.yml
@@ -37,7 +37,7 @@ jobs:
       # Build universal apk.
       - run: flutter build apk --debug --flavor github
       # Upload universal generated apk to the artifacts.
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: Musify-debug.apk
           path: build/app/outputs/flutter-apk/app-github-debug.apk
--- a/lib/screens/settings_page.dart
+++ b/lib/screens/settings_page.dart
@@ -230,23 +230,6 @@ class SettingsPage extends StatelessWidget {
           },
         ),
         ValueListenableBuilder<bool>(
-          valueListenable: playNextSongAutomatically,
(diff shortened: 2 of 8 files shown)


gokadzev:
@Shrawan13-glitch Hi, thank you for your contribution. 


I see too much changes here, so I gotta keep it on hold until the version 11.0. Before:



- It copies views, likes, etc. into the title, which is wrong. It should keep only the artist and song title.
- Should we start using the same image for high, low, and regular properties just because it doesn’t provide different resolution images?
- Could you please elaborate on what this rapid-skip issue is about?
- Was there a chance we could get the same result with less additions?

Shrawan13-glitch:
1. You’re right. I’ll fix that.
2. Using the same image for high/low/regular was only a fallback because the queue response didn’t always provide multiple thumbnail sizes. I understand that this should be handled better.
3. The rapid-skip issue happened when songs were skipped quickly using next/previous, for example 5-7 songs at a time. Intermediate tracks could briefly play before the final intended song started.
4. I think this can be done with fewer additions. I’ll close this for now and revisit it later in a smaller PR after version 11.0.
Thanks for reviewing.

PR #800, opened 2026-03-17
fix: offline fallback for youtube playlists

elias001011:
Changes:
- When offline mode is enabled, a downloaded playlist now loads directly from the offline list.
- Removed the online fallback/timeout; online mode keeps the original network flow.


elias001011:
Solved in the last commit

PR #809, opened 2026-03-24
Fix/add to queue and play next 

elias001011:
## Summary
Fixes the issue where songs added via "Play Next" or "Add to Queue" disappear when the queue is replaced or reordered.

## Problem
When users added songs to the queue using the "Play Next" or "Add to Queue" buttons, the songs would disappear if:
- A new playlist was played (replacing the queue)
- The shuffle mode was toggled

## Solution
Introduced an `isManuallyAdded` flag to track songs that were explicitly added by the user. This flag is set when songs are added via:
- `playNext()`
- `addToQueue()`

The implementation preserves manually added songs in three scenarios:

1. **Queue replacement** (`addPlaylistToQueue` with `replace=true`)
   - Before clearing the queue, extract all unplayed manually added songs
   - After populating the new queue, re-insert manually added songs after the currently playing song

2. **Shuffle activation**
   - Extract manually added songs before shuffling
   - Insert them back after the current song position (index 1) after shuffle completes

3. **Shuffle deactivation**
   - Extract manually added songs from the shuffled queue
   - Remove them from the restored original queue
   - Re-insert them after the current song position

## Behavior
- Manually added songs persist until they are played
- Songs are always inserted right after the currently playing song (index `_currentQueueIndex + 1`)
- This is slightly more permissive than Spotify's implementation (which only preserves songs when reordering), but provides better UX as users don't lose songs they intentionally queued

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -739,7 +739,9 @@ class MusifyAudioHandler extends BaseAudioHandler {
         insertIndex = _queueList.length;
       }
 
-      _queueList.insert(insertIndex, _queueEntryIds.createSong(song));
+      final queueSong = _queueEntryIds.createSong(song);
+      queueSong['isManuallyAdded'] = true;
+      _queueList.insert(insertIndex, queueSong);
 
       if (_currentQueueIndex < 0) {
         _currentQueueIndex = 0;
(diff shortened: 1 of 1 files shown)


gokadzev:
@elias001011 Hi, please rebase the branch, it includes one extra commit from me.

PR #830, opened 2026-04-25
fix: respect offline playlist sort order in player queue

elias001011:
## Description
This PR fixes a bug where the playback queue for "Offline Songs" would always follow the default order, ignoring any active sorting criteria (e.g., "Date Added", "Title", or "Artist").

## Changes
- Modified `UserSongsPage` to ensure that when a song is played (either via the main "Play" button or by tapping a specific song), the list passed to the `audioHandler` is correctly sorted according to the user's current selection.
- Used the existing `_sortOfflineSongsLocal` and `_getCurrentOfflineSortType` logic to derive the sorted list before initializing the playlist payload.
- This ensures the player queue matches the visual order of songs shown on the screen.

### DIFF
--- a/lib/screens/user_songs_page.dart
+++ b/lib/screens/user_songs_page.dart
@@ -206,11 +206,18 @@ class _UserSongsPageState extends State<UserSongsPage> {
                     label: Text(context.l10n!.play),
                     onPressed: () {
                       final songsList = getSongsList(widget.page);
+                      var sortedList = songsList;
+                      if (isOfflineSongs) {
+                        sortedList = _sortOfflineSongsLocal(
+                          songsList,
+                          _getCurrentOfflineSortType(),
+                        );
+                      }
                       final playlist = {
                         'ytid': '',
                         'title': title,
                         'source': 'user-created',
-                        'list': songsList,
+                        'list': sortedList,
                       };
                       audioHandler.playPlaylistSong(
                         playlist: playlist,
@@ -334,11 +341,18 @@ class _UserSongsPageState extends State<UserSongsPage> {
           builder: (_, searchQuery, __) {
             final isSearching = searchQuery.isNotEmpty;
             final displayList = _getDisplayList(songsList);
+            var sortedList = songsList;
+            if (isOfflineSongs) {
+              sortedList = _sortOfflineSongsLocal(
+                songsList,
+                _getCurrentOfflineSortType(),
+              );
+            }
             final playlist = {
               'ytid': '',
               'title': title,
               'source': 'user-created',
-              'list': songsList,
+              'list': sortedList,
             };
 
             if (displayList.isEmpty) {


gokadzev:
@elias001011 What's the situation with online playlists? Does it handle the case correctly?

elias001011:
@gokadzev Hi! I've checked the implementation details. In PlaylistPage, the sorting logic directly modifies the _playlist['list'] object, so the audioHandler receives the already sorted list.

However, in UserSongsPage (Offline Songs), the sorting is only applied to a local variable for display purposes (_getDisplayList), while the 'Play' button and song taps still reference the original unsorted userOfflineSongs list.

My PR fixes this specifically for the offline section by ensuring the sorted list is passed to the audioHandler, matching the behavior users expect from the UI

gokadzev:
> @gokadzev Hi! I've checked the implementation details. In PlaylistPage, the sorting logic directly modifies the _playlist['list'] object, so the audioHandler receives the already sorted list.
> 
> However, in UserSongsPage (Offline Songs), the sorting is only applied to a local variable for display purposes (_getDisplayList), while the 'Play' button and song taps still reference the original unsorted userOfflineSongs list.
> 
> My PR fixes this specifically for the offline section by ensuring the sorted list is passed to the audioHandler, matching the behavior users expect from the UI

UserSongsPage is used not only for offline songs, but also recently played songs and liked songs

elias001011:
> > @gokadzev Hi! I've checked the implementation details. In PlaylistPage, the sorting logic directly modifies the _playlist['list'] object, so the audioHandler receives the already sorted list.
> > 
> > However, in UserSongsPage (Offline Songs), the sorting is only applied to a local variable for display purposes (_getDisplayList), while the 'Play' button and song taps still reference the original unsorted userOfflineSongs list.
> > 
> > My PR fixes this specifically for the offline section by ensuring the sorted list is passed to the audioHandler, matching the behavior users expect from the UI
> 
> UserSongsPage is used not only for offline songs, but also recently played songs and liked songs

No, the other sections (liked songs and recently played) don't have any sorting feature at all - they're just displayed in the default order. So the bug only affects offline songs because that's the only section that has sorting options. My fix specifically targets only the offline songs case and won't have any effect on the other sections since they don't have sorting functionality.

gokadzev:
> > > @gokadzev Hi! I've checked the implementation details. In PlaylistPage, the sorting logic directly modifies the _playlist['list'] object, so the audioHandler receives the already sorted list.
> > > However, in UserSongsPage (Offline Songs), the sorting is only applied to a local variable for display purposes (_getDisplayList), while the 'Play' button and song taps still reference the original unsorted userOfflineSongs list.
> > > My PR fixes this specifically for the offline section by ensuring the sorted list is passed to the audioHandler, matching the behavior users expect from the UI
> > 
> > 
> > UserSongsPage is used not only for offline songs, but also recently played songs and liked songs
> 
> No, the other sections (liked songs and recently played) don't have any sorting feature at all - they're just displayed in the default order. So the bug only affects offline songs because that's the only section that has sorting options. My fix specifically targets only the offline songs case and won't have any effect on the other sections since they don't have sorting functionality.

That's what I was trying to recall. Approved then. Thanks!

PR #848, opened 2026-05-29
perf: reduce startup blocking work

elias001011:
## Summary

- Defer the F-Droid announcement fetch until after the first frame.
- Cache Home page futures to avoid refetching suggested playlists/recommendations on unrelated rebuilds.
- Refresh the cached liked-playlists future when the liked playlist count changes.
- Create offline directories in parallel without blocking app startup, with error logging for background failures.

## Why

This keeps the PR focused on startup/performance only. Media-resumption related changes were removed from this PR and split into #851.

The incoming app-links listener is still registered during initialization, because `app_links` recommends subscribing early enough to catch the first link. Only non-critical post-startup work is deferred.

## Testing

- `flutter analyze`
- `flutter build apk --debug --flavor github -t lib/main.dart`

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -249,6 +249,9 @@ void main() async {
   await initialisation();
 
   runApp(const Musify());
+  SchedulerBinding.instance.addPostFrameCallback((_) {
+    unawaited(postRunInitialisation());
+  });
 }
 
 Future<void> initialisation() async {
@@ -287,16 +290,37 @@ Future<void> initialisation() async {
     } on PlatformException {
       logger.log('Failed to get initial uri');
     }
-
-    if (isFdroidBuild && !offlineMode.value) {
-      await fetchAnnouncementOnly();
-    }
   } catch (e, stackTrace) {
     logger.log('Initialization Error', error: e, stackTrace: stackTrace);
   }
 
   applicationDirPath = (await getApplicationDocumentsDirectory()).path;
-  await FilePaths.ensureDirectoriesExist();
+  unawaited(
+    FilePaths.ensureDirectoriesExist().catchError((
+      Object error,
+      StackTrace stackTrace,
+    ) {
+      logger.log(
+        'Failed to create application directories',
+        error: error,
+        stackTrace: stackTrace,
+      );
+    }),
+  );
+}
+
+Future<void> postRunInitialisation() async {
+  try {
+    if (isFdroidBuild && !offlineMode.value) {
+      await fetchAnnouncementOnly();
+    }
+  } catch (e, stackTrace) {
+    logger.log(
+      'Post-run initialization error',
+      error: e,
+      stackTrace: stackTrace,
+    );
+  }
 }
 
 void handleIncomingLink(Uri? uri) async {
--- a/lib/screens/home_page.dart
+++ b/lib/screens/home_page.dart
@@ -43,6 +43,52 @@ class HomePage extends StatefulWidget {
 }
 
 class _HomePageState extends State<HomePage> {
+  late Future<List<dynamic>> _suggestedPlaylistsFuture;
+  late Future<List<dynamic>> _likedPlaylistsFuture;
+  late Future<List<dynamic>> _recommendedSongsFuture;
+
+  @override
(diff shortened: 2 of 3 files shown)


gokadzev:
@elias001011 It's a bit unclear for me in which cases it's possible that we can access/click play button but don't have any active media in the app (if it's about notification, when the app is closed and notification is kept, listener and ongoing services are killed anyway)

elias001011:
> @elias001011 It's a bit unclear for me in which cases it's possible that we can access/click play button but don't have any active media in the app (if it's about notification, when the app is closed and notification is kept, listener and ongoing services are killed anyway)

Thanks for checking it.

This is not about Musify's own playback notification being kept after the app is closed. I mean Android's system media resumption/player widget, the one shown in Quick Settings / media controls.

On Android, media apps can expose a recent playable item through MediaBrowserService. Then, even after the app/service is gone, Android can still show the app as a recent media source. When the user taps play there, Android starts the app/service again and asks it to prepare/play that recent media item.

This is the behavior I was comparing with Spotify: after playing something, Spotify remains available as a recent media app in the system media widget, and tapping it brings Spotify back. Before this PR, Musify often did not appear there as a resumable recent player, almost as if Android did not know Musify had been the last media app.

So the PR is not trying to control playback from a dead notification listener. It exposes a lightweight recent media item and implements `getChildren`, `getMediaItem`, `prepareFromMediaId`, and `playFromMediaId` so Android media resumption has something valid to display and request when the user taps the system widget.

I agree this part is a bit sensitive and Android-version-dependent, so review is very welcome. In my local testing it improved system media integration without affecting normal playback.

gokadzev:
> > @elias001011 It's a bit unclear for me in which cases it's possible that we can access/click play button but don't have any active media in the app (if it's about notification, when the app is closed and notification is kept, listener and ongoing services are killed anyway)
> 
> Thanks for checking it.
> 
> This is not about Musify's own playback notification being kept after the app is closed. I mean Android's system media resumption/player widget, the one shown in Quick Settings / media controls.
> 
> On Android, media apps can expose a recent playable item through MediaBrowserService. Then, even after the app/service is gone, Android can still show the app as a recent media source. When the user taps play there, Android starts the app/service again and asks it to prepare/play that recent media item.
> 
> This is the behavior I was comparing with Spotify: after playing something, Spotify remains available as a recent media app in the system media widget, and tapping it brings Spotify back. Before this PR, Musify often did not appear there as a resumable recent player, almost as if Android did not know Musify had been the last media app.
> 
> So the PR is not trying to control playback from a dead notification listener. It exposes a lightweight recent media item and implements `getChildren`, `getMediaItem`, `prepareFromMediaId`, and `playFromMediaId` so Android media resumption has something valid to display and request when the user taps the system widget.
> 
> I agree this part is a bit sensitive and Android-version-dependent, so review is very welcome. In my local testing it improved system media integration without affecting normal playback.

Interesting, is there any chance you could move media-resumption related changes into a separate PR?

elias001011:
> Interesting, is there any chance you could move media-resumption related changes into a separate PR?

Sure

elias001011:
You’re right, that part deserves to be reviewed separately. I moved the media-resumption related changes out of this PR and force-pushed this branch so #848 now only contains the startup/performance changes.

I kept the media-resumption work on a separate local branch for now and can open a focused PR for it after we have clearer reproduction/testing around the Samsung media widget behavior.

elias001011:
Opened the media-resumption work separately as draft PR #851: https://github.com/gokadzev/Musify/pull/851

This PR now remains focused only on startup/performance changes.

elias001011:
I did one more pass to reduce the behavioral risk of this PR:

- rebased it on the latest `master`
- kept the `app_links` stream subscription in initialization instead of deferring it, since the package recommends subscribing early enough to catch the first link
- kept only the F-Droid announcement fetch as post-frame work
- added error logging for the background directory creation
- refreshed the cached liked-playlists future when liked playlists change, so the Home cache does not go stale

Retested with:

- `flutter analyze`
- `flutter build apk --debug --flavor github -t lib/main.dart`

PR #851, opened 2026-05-29
feat: support Android media resumption

elias001011:
## Summary

- Expose a recent playable media item through AudioService/MediaBrowserService.
- Implement getChildren/getMediaItem/prepareFromMediaId/playFromMediaId for Android media resumption.
- Allow Android’s system media widget to request a recent item when it tries to resume Musify.

## Context

This was originally reported on a Samsung device where Musify did not behave like a resumable recent media app in Android’s system media widget. On my Xiaomi device, Musify may already be recognized without this change, so this PR is intentionally separated and kept as draft for focused review and additional testing across devices/Android vendor builds.

Split out from #848 after maintainer feedback.

## Review updates

- Added active queue lookup so media ids from the current queue can be resolved before falling back to persisted lists.
- Tightened resumable song normalization so media resumption only exposes songs with a valid ytid, keeping id/ytid string handling consistent.

## Testing

- `flutter analyze`
- `flutter build apk --debug --flavor github -t lib/main.dart`

## Status

Draft: I plan to test this more carefully on multiple Android devices/vendor builds before marking it ready.

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -97,6 +97,7 @@ class MusifyAudioHandler extends BaseAudioHandler {
   static const Duration _errorRetryDelay = Duration(seconds: 2);
   static const Duration _songTransitionTimeout = Duration(seconds: 30);
   static const Duration _debounceInterval = Duration(milliseconds: 150);
+  static const String _recentMediaIdPrefix = 'recent:';
 
   int _activePreloadCount = 0;
   final Set<String> _preloadingYtIds = <String>{};
@@ -752,7 +753,11 @@ class MusifyAudioHandler extends BaseAudioHandler {
         await _playFromQueue(0);
       }
     } catch (e, stackTrace) {
-      logger.log('Error inserting recommended song', error: e, stackTrace: stackTrace);
+      logger.log(
+        'Error inserting recommended song',
+        error: e,
+        stackTrace: stackTrace,
+      );
     }
   }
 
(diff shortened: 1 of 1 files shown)


lib/services/audio_service.dart
```diff
@@ -1256,6 +1261,168 @@ class MusifyAudioHandler extends BaseAudioHandler {
...
+        _firstPlayableSong(userLikedSongsList);
+  }
+
+  Map<String, dynamic> _normaliseResumableSong(Map song) {
+    final normalised = cloneMap(song);
+    final ytid = normalised['ytid']?.toString();
+    if (ytid != null && ytid.isNotEmpty) {
+      normalised['id'] = ytid;
+    } else {
+      normalised['id'] ??= 0;
```
gokadzev: if ytid has something assigned it's string type, but if it doesn't, we assign a number?

lib/services/audio_service.dart
```diff
@@ -1256,6 +1261,168 @@ class MusifyAudioHandler extends BaseAudioHandler {
...
+
+  Map? _findSongByYtid(String? ytid) {
+    if (ytid == null || ytid.isEmpty) return null;
+
+    final activeSong = currentSong;
+    if (activeSong?['ytid']?.toString() == ytid) {
+      return activeSong;
+    }
+
+    for (final source in [
```
gokadzev, line 1304: it might be a good idea to search it in an active queue as well (_queueList)

lib/services/audio_service.dart
elias001011: Good catch. The earlier fallback was too loose: for this resumption path we can only safely expose/play an item if it has a valid `ytid`, because that is the stable id Musify can resolve back into a playable song. I changed the normalization so it returns no resumable item without `ytid`, and when `ytid` exists both `id` and `ytid` are kept as that string value.

lib/services/audio_service.dart
elias001011: Agreed. A SystemUI resumption request can refer to something from the active playback session before it is necessarily present in recently played/offline/liked lists. I added `_queueList` to the lookup, after `currentSong` and before the persisted lists, so active queue entries can be resolved as well.

elias001011:
Tested and validated on several devices, including samsung and in fact, in the case of samsung, we had an improvement in the integration with the widget, being able to play the last song directly there. I'll leave this PR ready for review in case you decide to merge

PR #852, opened 2026-05-30
fix: prevent duplicate liked items

elias001011:
## Summary
- prevent rapid liked song toggles from creating duplicate entries
- centralize liked song and liked playlist count updates in their services
- deduplicate liked songs/playlists by `ytid` before publishing and persisting state
- pass already-loaded song/playlist data into like status updates to avoid unnecessary fallback fetches

## Root cause
Like status updates could be triggered several times before async metadata fetches completed. Each stale completion could append another liked item or manually nudge the count notifier from the UI, leaving the stored list and listeners out of sync.

## Tests
- `flutter analyze`
- `git diff --check`
- manual: rapid like/unlike on liked songs no longer creates duplicate liked song entries
- manual: playlist like/unlike still updates the Library liked playlists section correctly

Note: `flutter test` currently fails on the existing template `test/widget_test.dart` because Hive boxes are not opened before mounting `Musify()`.

### DIFF
--- a/lib/screens/playlist_page.dart
+++ b/lib/screens/playlist_page.dart
@@ -369,10 +369,9 @@ class _PlaylistPageState extends State<PlaylistPage> {
                     updatePlaylistLikeStatus(
                       _playlist['ytid'],
                       playlistLikeStatus.value,
+                      playlistData: _playlist,
                     ),
                   );
-                  currentLikedPlaylistsLength.value =
-                      currentLikedPlaylistsLength.value - 1;
                 },
                 tooltip: context.l10n!.removeFromLikedSongs,
               )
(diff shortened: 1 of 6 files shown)


elias001011:
I opened this as a proposed fix for the liked item state/update issue we discussed.

The PR ended up covering two related flows:

- liked playlists: removes the manual UI increment/decrement after `updatePlaylistLikeStatus()`, so the service remains the single place updating `currentLikedPlaylistsLength`.
- liked songs: while testing rapid like/unlike taps, I could reproduce duplicated entries in Liked Songs, so this also adds protection there by deduplicating by `ytid` and ignoring stale async like updates.

This may be bigger than the minimum change needed for the original playlist notifier issue. Could you take a look and tell me whether you think all of this belongs in one PR, or if you'd prefer I trim it down to only the playlist notifier cleanup and handle the liked songs duplicate case separately?

PR #853, opened 2026-05-30
Fix content hidden behind floating mini player

elias001011:
## Summary

Fixes scrollable content being hidden behind the floating mini player.

The mini player stays floating and transparent, while affected scrollable pages get enough bottom scroll extent for the last item to move above it.

## Details

- Keeps the mini player in the existing `Stack`, preserving the floating transparent appearance.
- Keeps the single `audioHandler.mediaItem` subscription in `BottomNavigationPage` for the inherited bottom padding used by menus and scroll safe areas.
- Uses shared native `SafeArea` / `SliverSafeArea` constants at the end of affected scroll views instead of custom spacer widgets.
- Uses the full mini-player footprint (`miniPlayerTotalHeight`) so the bottom inset includes the player and floating margin.

## Why this approach

Root-level Scaffold/body padding changes can stop content from painting behind the transparent mini player area. This keeps the visual behavior intact and only extends each scroll view at the end, where the extra space is needed.

This version also avoids per-page listeners and avoids custom inherited/spacer widgets.

## Validation

- `flutter analyze`
- `git diff --check`

`flutter test` still fails on the existing template smoke test because Hive boxes are not opened before `Musify` is pumped (`HiveError: Box not found`). That is unrelated to this UI change.

### DIFF
--- a/lib/screens/about_page.dart
+++ b/lib/screens/about_page.dart
@@ -25,6 +25,7 @@ import 'package:musify/constants/app_constants.dart';
 import 'package:musify/constants/version.dart';
 import 'package:musify/extensions/l10n.dart';
 import 'package:musify/utilities/url_launcher.dart';
+import 'package:musify/widgets/mini_player_bottom_space.dart';
 
 class AboutPage extends StatelessWidget {
   const AboutPage({super.key});
@@ -159,6 +160,7 @@ class AboutPage extends StatelessWidget {
                 ),
               ),
             ),
+            const MiniPlayerBottomSpace(),
           ],
         ),
       ),
--- a/lib/screens/bottom_navigation_page.dart
+++ b/lib/screens/bottom_navigation_page.dart
@@ -24,6 +24,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
 import 'package:go_router/go_router.dart';
+import 'package:musify/constants/app_constants.dart';
 import 'package:musify/extensions/l10n.dart';
 import 'package:musify/main.dart';
 import 'package:musify/services/settings_manager.dart';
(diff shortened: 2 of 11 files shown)


elias001011:
Tried reducing the PR.

The custom spacer/inherited widget approach is gone now. The PR keeps the single media item listener in `BottomNavigationPage` and uses native `SafeArea` / `SliverSafeArea` at the end of affected scroll views, so the floating transparent mini player behavior stays the same with a much smaller diff.

`flutter analyze` and `git diff --check` pass.

elias001011:
I pushed one small follow-up commit separately.

I noticed the bottom padding only depends on whether the mini player exists, not on the current song metadata. So `BottomNavigationPage` now listens to mini player visibility (`mediaItem != null`) with `.distinct()` instead of the full `mediaItem` stream.

That means changing tracks still updates the mini player itself, but it should not rebuild the page padding behind it unless the mini player actually appears or disappears.

elias001011:
Ok, I tried a different approach and updated the PR.

This version combines the reusable base screen idea with the previous scroll-inset approach: the extra mini player space is still added inside the scrollable area, but now it is handled by reusable `MiniPlayerScrollableScaffold`, `MiniPlayerListScaffold`, and `MiniPlayerSliverScaffold` wrappers instead of manually adding a padding/safe-area widget on every screen.

The only downside is that the diff got larger because several screens now use the new wrappers. I kept the commits separated so you can look at what changed and compare the different approaches more easily.

elias001011:
I tried the reusable scaffold approach, but after testing it I think it goes too far for this fix.

It technically works, but it makes the diff much larger and does not really remove the core tradeoff: because the mini player is transparent/floating, the extra space still has to be applied inside the scrollable area. The scaffold wrappers mostly moved that detail around and reworked several screens, which feels like too much churn for a mini player inset.

I reverted the PR back to the smaller approach. It is still manual per affected scroll view, but the final diff is small and explicit, and I think that is easier to review and maintain here.

elias001011:
I restored the reusable scaffold commit so you can compare both options directly in the PR.

The first two commits are the smaller explicit approach. The last commit is the experiment that combines that behavior with reusable scroll scaffolds, so the bottom safe area is handled by shared wrappers instead of being added manually in each screen.

The tradeoff is pretty clear: the scaffold version removes the repeated manual widget, but it also makes the diff much larger because several screens need to be converted to the wrappers. I kept it as a separate commit so it is easy to review or drop depending on which direction you prefer.

Issue #826, opened 2026-04-12
[Feature] Monthly Listening Retrospective / Wrapped


Charlie-codev:
### Is there an existing issue for this?

- [x] I have searched the existing issues and confirmed that this feature is not already requested.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this feature.

### Credits

me

### Pain

Currently, Musify only shows recent listening history, but users have no way to track their listening patterns over time in a meaningful way. There's no visualization of monthly listening habits or insights about which songs dominated the user's experience throughout the year.

The lack of a retrospective feature means users miss out on the satisfying experience of discovering their "anthem" for each month or seeing how their music taste evolved. This is a standard feature that music streaming apps offer (like Spotify Wrapped), and its absence creates a gap in the user experience, especially for users who use Musify as their primary music app and want to feel more connected to their listening journey.

Users also cannot easily identify patterns in their listening behavior, such as seasonal variations in music taste or which genres/artists they gravitated toward during specific periods.

### Feature Overview

Implement a **Monthly Listening Retrospective** feature that provides users with insights about their music listening habits on a monthly and annual basis.

### Implementation Details

1. **Data Tracking System**
   - Track play count for each song with timestamps
   - Store monthly aggregates (total plays per song per month)
   - Calculate the "Top Song of the Month" based on play count
   - Maintain annual statistics for year-end summaries

2. **UI Placement Options**
   - A dedicated banner on the Home tab (collapsible after viewed)
   - A section in Settings called "Your Music Year"
   - A notification prompt when new monthly data is available
   - The banner should display the "Song of the Month" with album art

3. **Retrospective Page**
   - Show the user's most played song of the current month
   - Display top 10 songs per month with play counts
   - Allow navigation through all months of the current year
   - Show a visual timeline or calendar-style view
   - Include the annual summary showing top songs of the year
   - Each month card should be expandable to show full top 10 list

4. **User Experience**
   - Reset data at the start of each new year
   - Option to share monthly/annual statistics
   - Include listening time statistics if feasible
   - Smooth animations for transitioning between months
   - Offline access to previously viewed retrospectives

5. **Technical Considerations**
   - Leverage existing play history data structure
   - Consider local database for storing statistics
   - Privacy-focused: all data stays on device
   - Consider export functionality for user data

### Useful Resources/References

[...]

PR #854, opened 2026-06-04
Add Time Machine listening recaps

elias001011:
#826

## Summary
- Adds Time Machine monthly and annual listening recaps with the Home current-month entry replacing the old Most Played surface.
- Keeps Wrapped metrics in the user Hive box under wrappedListeningStats, separate from Recently Played. Recently Played remains for UI and recommendations and is capped at 100 items.
- Tracks listening time on playback ticks while stats are enabled; ranked Wrapped plays are only counted after the qualified threshold: 30 seconds, or half the duration for very short tracks.
- Stores up to 250 ranked songs for the active month, archives top 15 per finished month, and builds the annual recap by aggregating months for the recap year.
- Renders Time Machine recap sections lazily so year-heavy histories do not build every monthly card and artwork thumbnail on entry.
- Keeps recap UI as a snapshot instead of listening for real-time stats updates, avoiding unnecessary rebuilds while playback ticks are persisted.
- Normalizes restored stats so stale currentMonth data moves into the correct history month, old Most Played data is ignored, and previous-year history is removed after January.
- Polishes recap UI: no debug data, no 0-minute cards, long-press actions on ranked songs, corrected artwork handling, and annual recap only in Time Machine.
- Uses share_plus for recap image sharing and includes the Musify icon asset for the rendered/shared recap header.

## Validation
- flutter analyze
- flutter build apk --debug --flavor github

## Notes
- No test files are included in this PR. The repo CI currently does not run flutter test, so the extra local utility tests were removed to keep the review diff focused.
- Kept the PR as draft until Liam and Valeri decide to mark it ready.


lib/services/router_service.dart
```diff
@@ -76,6 +78,10 @@ class NavigationManager {
           return homePath;
         }
 
+        if (isOffline && currentPath == timeMachinePath) {
```
gokadzev: this can be combined with above one (if  current path is searchpath or timemachinepath

pubspec.yaml
```diff
@@ -35,6 +35,7 @@ dependencies:
   receive_sharing_intent:
     path: ./packages/receive_sharing_intent
   rxdart: ^0.28.0
+  share_plus: ^12.0.2
```
gokadzev, line 38: unused dependency

pubspec.yaml
```diff
@@ -52,6 +53,7 @@ flutter:
   # To add assets to your application, add an assets section, like this:
   assets:
     - assets/licenses/
+    - android/app/src/main/res/drawable-xxxhdpi/ic_launcher_foreground.png
```
gokadzev: what's this for?

lib/services/router_service.dart
elias001011: Done, combined both offline redirects into one condition.

pubspec.yaml
elias001011: It is used in TimeMachinePage to share the generated recap image via SharePlus.instance.share.

pubspec.yaml
elias001011: This is the existing Musify launcher foreground icon. I use it as the small brand icon in the recap header that gets rendered/shared, so it is listed as an asset for AssetImage.

lib/services/router_service.dart
```diff
@@ -147,6 +150,11 @@ class NavigationManager {
               );
             },
             routes: [
+              GoRoute(
+                path: 'time-machine',
```
gokadzev: camelCase would work better I guess, as we use it in most of the places (timeMachine)

lib/services/audio_service.dart
```diff
@@ -238,6 +249,15 @@ class MusifyAudioHandler extends BaseAudioHandler {
     });
   }
 
+  void _startAudioPlayer() {
```
gokadzev: what's this for?

lib/widgets/song_bar.dart
```diff
@@ -268,58 +613,18 @@ class _SongBarState extends State<SongBar> {
   }
 
   void _handleMenuAction(BuildContext context, String value) {
```
gokadzev, line 647: Function to call another function?

lib/services/router_service.dart
elias001011: Done. I switched the route to `timeMachine` in the router and updated the Home button to match. I also ran `flutter analyze` on the touched files and it is clean.

lib/services/audio_service.dart
elias001011: Its a small wrapper around audioPlayer.play() so playback can start without blocking the rest of the flow and the error logging stays centralized. I use it in both the resume path and the normal track-start path, and the listening stats session begins right after the play request.

lib/widgets/song_bar.dart
elias001011: Yes, intentionally. I extracted the menu logic so SongBar and the recap cards can share it, and this wrapper just forwards the widget-specific state into the shared handler.

lib/services/audio_service.dart
gokadzev: I think we're making the codebase unreasonably big, it's just a one call, no need to have separate function for it.

elias001011: Fair point. I removed `_startAudioPlayer()` and inlined the same `audioPlayer.play()` call in both places, so the flow stays the same without the extra abstraction.

lib/localization/app_en.arb
gokadzev, line 1: you could add new translations here and ignore other files btw

lib/screens/home_page.dart
```diff
@@ -217,65 +221,101 @@ class _HomePageState extends State<HomePage> {
     );
   }
 
-  Widget _buildMostPlayedSection() {
-    final sectionTitle = context.l10n!.mostPlayed;
+  Widget _buildCurrentMonthRecapSection() {
+    return ValueListenableBuilder<bool>(
+      valueListenable: offlineMode,
```
gokadzev: is valuelistenable necessary here? we completely remove home page and replace it with offline songs playlist on offline mode so I believe it's not possible to visit an actual home page while having enabled offline mode

lib/screens/home_page.dart
```diff
@@ -217,65 +221,101 @@ class _HomePageState extends State<HomePage> {
...
+    int index,
+  ) async {
+    if (songs.isEmpty) return;
+    await audioHandler.playPlaylistSong(
+      playlist: {'title': context.l10n!.timeMachine, 'list': songs},
+      songIndex: index,
+    );
+  }
+
+  String _formatMonthPeriodLabel(BuildContext context, String monthKey) {
```
gokadzev: let's pass Locale directly, instead of passing the whole context. also you can consider moving this helper in utils

lib/services/audio_service.dart
```diff
@@ -89,6 +91,14 @@ class MusifyAudioHandler extends BaseAudioHandler {
   bool _completionEventPending = false;
   bool _completionHandlerLoadStarted = false;
 
+  Map? _listeningStatsSong;
+  String? _listeningStatsSongId;
+  Duration? _listeningStatsDuration;
+  Duration _listeningStatsListened = Duration.zero;
+  DateTime? _listeningStatsLastTick;
+  bool _listeningStatsQualified = false;
+  bool _lastAudioPlayerPlaying = false;
```
gokadzev: we can move these variables along with listening recaps-related functions in listening_stats_service

lib/localization/app_en.arb
elias001011: Good point — added the new translation keys only in `app_en.arb` and left the other locale files untouched, so translators can pick them up via the normal workflow.

lib/screens/home_page.dart
elias001011: You are right — the home route is replaced entirely by the offline songs playlist when offline mode is active, so `HomePage` is never reached in that state. Removed the `offlineMode` `ValueListenableBuilder` wrapper; `wrappedEnabled` is the only guard needed here.

lib/screens/home_page.dart
elias001011: Done — `_formatMonthPeriodLabel` is now a top-level `formatMonthPeriodLabel(Locale, String)` helper in `app_utils.dart` (accepts `Locale` directly, not `BuildContext`). The call site in `HomePage` was updated accordingly and the `intl` import was moved to `app_utils.dart` where it belongs.

lib/services/audio_service.dart
elias001011: Done — all session-tracking state (`_sessionSong`, `_sessionSongId`, `_sessionDuration`, `_sessionListened`, `_sessionLastTick`, `_sessionQualified`, `_sessionLastAudioPlayerPlaying`) and the corresponding lifecycle methods (`startListeningSession`, `resumeListeningSession`, `recordListeningSessionProgress`, `handlePlayerStateForListeningStats`, `finishListeningSession`, etc.) have been moved into `ListeningStatsService`. `MusifyAudioHandler` now just delegates to `listeningStatsService.*` for every stats-related event, keeping the handler focused on transport only.

elias001011:
**Latest changes (2026-06-17):**

- Removed all new translation strings from the non-EN locale files — only `app_en.arb` now carries the new keys, as suggested. The other locale files are restored to their upstream state.
- Fixed a pre-existing `unawaited_futures` lint warning in `song_bar.dart` (the `context.push()` call in the "Go to Artist" menu action was missing `unawaited`). Not related to this feature, just a `flutter analyze` cleanup.

elias001011:
Wrapped/time machine refactor summary: I moved listening stats out of the hot path so we no longer write to Hive every 10 seconds. The service now accumulates listening time in memory and flushes it at meaningful checkpoints such as track changes, pause/stop/seek, app backgrounding, and playback completion.\n\nI also validated the main flows locally: the app still accumulates listened time correctly, preserves the final tick when a track changes, and keeps persistence working without the periodic Hive writes.

lib/services/listening_stats_service.dart
```diff
@@ -0,0 +1,509 @@
...
+
+  void resetListeningSession({
+    bool countCurrentTick = false,
+    bool flushStats = true,
+  }) {
+    finishListeningSession(
+      countCurrentTick: countCurrentTick,
+      flushStats: flushStats,
+    );
+  }
```
gokadzev: redundant function

lib/services/listening_stats_service.dart
```diff
@@ -0,0 +1,509 @@
...
+    if (!wrappedEnabled.value) return;
+
+    final song = currentSong;
+    if (song == null) return;
+
+    final ytid = song['ytid']?.toString();
+    if (ytid == null || ytid.isEmpty) return;
+
+    if (_sessionSongId != ytid) {
+      finishListeningSession(countCurrentTick: true, wasPlaying: true);
```
gokadzev, line 201: why is wasPlaying hardcoded here?

lib/services/listening_stats_service.dart
```diff
@@ -0,0 +1,509 @@
...
+    if (song == null) return;
+
+    final now = DateTime.now();
+    final lastTick = _sessionLastTick;
+    _sessionLastTick = now;
+    if (lastTick == null) return;
+
+    final shouldCount =
+        (wasPlaying ?? _sessionLastAudioPlayerPlaying) ||
+        (force && _sessionLastAudioPlayerPlaying);
```
gokadzev: Does force actually do anything that _sessionLastAudioPlayerPlaying and wasPlaying don't?

lib/services/listening_stats_service.dart
```diff
@@ -0,0 +1,509 @@
...
+  String? _sessionSongId;
+  Duration? _sessionDuration;
+  Duration _sessionListened = Duration.zero;
+  DateTime? _sessionLastTick;
+  bool _sessionQualified = false;
+  bool _sessionLastAudioPlayerPlaying = false;
+
+  bool get hasStats {
+    final now = DateTime.now();
+    final stats = _readStats();
```
gokadzev: why don't we pass now variable here too?

lib/utilities/listening_stats_utils.dart
```diff
@@ -0,0 +1,671 @@
...
+
+Map<String, dynamic> createEmptyListeningStats(
+  int year, {
+  String? currentMonthKey,
+}) {
+  final monthKey = currentMonthKey ?? listeningStatsMonthKey(DateTime(year));
+
+  return {
+    'schemaVersion': wrappedListeningStatsSchemaVersion,
+    'year': _parseMonthKey(monthKey)?.year ?? year,
```
gokadzev: why do we use this here like this but everywhere else it's just now.year?

lib/utilities/listening_stats_utils.dart
```diff
@@ -0,0 +1,671 @@
...
+      source['artworkPath'],
+      fallback: target['artworkPath'],
+    )
+    ..['artWorkPath'] = _readString(
+      source['artWorkPath'],
+      fallback: target['artWorkPath'],
+    )
+    ..['lastPlayed'] = source['lastPlayed'];
+
+  if (source['duration'] != null) target['duration'] = source['duration'];
```
gokadzev: no usage of _readPersistableDuration here?

lib/utilities/listening_stats_utils.dart
```diff
@@ -0,0 +1,671 @@
...
+  if (year == null || month == null) return null;
+  if (month < DateTime.january || month > DateTime.december) return null;
+
+  return DateTime(year, month);
+}
+
+Map<String, dynamic>? _asMap(dynamic value) {
+  if (value is! Map) return null;
+  return Map<String, dynamic>.from(value);
+}
```
gokadzev: _asMap and _normalizeMap look quite similar, couldn't we reuse normalizeMap instead of creating _asMap?

lib/widgets/song_bar.dart
```diff
@@ -39,6 +40,363 @@ import 'package:musify/utilities/playlist_dialogs.dart';
...
+  return RelativeRect.fromRect(
+    Rect.fromLTWH(position.dx, position.dy, 1, 1),
+    Offset.zero & size,
+  );
+}
+
+String _readSongYtid(dynamic song) =>
+    song is Map ? song['ytid']?.toString() ?? '' : '';
+
+List<PopupMenuEntry<String>> _buildSongMenuItems({
```
gokadzev, line 113: are you sure moving those as separate functions make sense? they are not called until user opens dropdown menu anyway

lib/services/listening_stats_service.dart
elias001011: Good catch — `resetListeningSession` was just a thin alias of `finishListeningSession`. Dropped it and call `finishListeningSession` directly from the audio handler's `resetListeningStatsSession`.

lib/services/listening_stats_service.dart
elias001011: It's intentional: this branch runs from `handlePlayerStateForListeningStats` *before* `_sessionLastAudioPlayerPlaying` is updated, so the flag still holds the pre-transition value (`false`). We just received a 'playing' event for a different song, meaning the previous one was playing up to this tick — using the stale flag would drop its final tick. Added a comment making the invariant explicit.

lib/services/listening_stats_service.dart
elias001011: You're right — in the state-change path `force` never changes the result, and it only mattered in `finishListeningSession` when `wasPlaying==false` while the cached flag was still `true` (a one-event race on stop/track-swap). Dropped it; the gate is now `wasPlaying ?? _sessionLastAudioPlayerPlaying`. The only behavioral diff is not counting a sub-second remainder in that rare race, negligible for the recap.

lib/services/listening_stats_service.dart
elias001011: Agreed, oversight — passing `now` keeps normalization and the annual-window check on a single timestamp. Fixed.

lib/utilities/listening_stats_utils.dart
elias001011: Simplified to `'year': year` — across all callers the `currentMonthKey`'s year always equals `year`, so the defensive `_parseMonthKey` round-trip was redundant (and the field isn't read by the annual-window logic anyway).

lib/utilities/listening_stats_utils.dart
elias001011: Good point — `_upsertSongStats` already normalizes via `_readPersistableDuration`; routed this copy path through it too. In practice the source is already-normalized stats, so it's defensive, but now consistent.

lib/utilities/listening_stats_utils.dart
elias001011: Right — `_asMap` was only feeding `monthTotalSeconds` (which accepts a nullable map and treats null/empty the same), so `_normalizeMap` covers it. Replaced both call sites and removed the util `_asMap`. (The service has its own `_asMap`; left it as-is unless you'd like it unified too.)

lib/widgets/song_bar.dart
elias001011: They're shared by two call sites (the long-press menu and the popup-menu button), so the extraction is about reuse/testability and avoiding capturing the `State` in closures — not perf. Happy to inline them or make them static methods if you'd prefer a different style, but functionally it keeps the two menus in sync.

PR #860, opened 2026-06-06
Fix Home recommendation reshuffling

elias001011:
## Summary
- Cache Home suggested playlist and recommended song futures in HomePage state instead of recreating them from build.
- Refresh recommended songs only when the external recommendations setting changes.

## Root cause
Home passed getPlaylists(...) and getRecommendedSongs() directly into AsyncLoader during build. Any rebuild could create fresh futures and rerun the shuffle logic while the page was still open.

## Validation
- flutter analyze
- flutter test currently fails in the existing smoke test before reaching Home because Hive boxes are not opened: `HiveError: Box not found. Did you forget to call Hive.openBox()?`


elias001011:
Refactored to remove manual listeners. Now using a dedicated _RecommendedSongsLoader widget that handles Future caching and updates via didUpdateWidget. This keeps the code declarative and aligned with the project's state management style while fixing the reshuffling bug.

gokadzev:
I'm conflicted about this PR. I don't think it has a significant impact on performance and it also generates updated recommendations each time. It's better to make code cleanups and optimizations in other places.

elias001011:
What I meant is: this wrapper keeps the `Future` stable so the recommendations do not refetch every time `HomePage` rebuilds, which is what was causing the visible jump/reshuffle. It only refreshes when `useExternalRecommendations` changes, since that is the case where the data source itself changes. If you prefer, I can still look for a more minimal fix that keeps the same behavior without this wrapper.

elias001011:
https://github.com/user-attachments/assets/88ea8318-4d8e-4592-82be-2fa08cf3c48c

gokadzev:
> Screenrecorder-2026-06-16-14-45-57-917.mp4

Oh, I didn't notice it, I wonder if it's caused by recent changes? also isn't it possible to solve it with more minimal and more efficient solution?

elias001011:
I went back to the earlier, more minimal approach from `6ad6c89f`. Do you think that version is better?

PR #869, opened 2026-06-12
Add YouTube Music artist pages

elias001011:
## Summary

Adds Artist Pages backed by a strict YouTube Music catalog source instead of regular channel uploads or Topic upload feeds.

Channel uploads were noisy and Topic uploads were clean but incomplete. This version resolves artists through YouTube Music's canonical artist search, then builds the catalog from the artist's official releases and tracks. If a canonical Music artist cannot be resolved, the Artist Page shows `Not found` and logs the lookup instead of opening a fake/noisy artist page.

## What changed

- Adds a separate local `youtube_music_explode_dart` package for the small YouTube Music API surface used by Musify.
- Keeps `youtube_explode_dart` separate; it is still used for normal YouTube playback/search metadata and stream handling.
- Centralizes artist search, resolution, catalog loading, thumbnail normalization, cache invalidation, and deduplication in `artist_service.dart`.
- Artist search now uses YouTube Music artist results, so queries can resolve to canonical artist names/images instead of echoing the typed text.
- Artist Pages open through the existing PlaylistPage behavior, preserving play, shuffle, like, add, sync/update, and download controls.
- Downloaded artists remain offline-first; sync/update forces a fresh Music catalog fetch when online.
- Search keeps verified/canonical artists mixed into song results and falls back to artist-based song search if the first song query returns empty.
- Invalid or mismatched artist names resolve to `Not found` and log lookup/source details.

## Validation

- `flutter analyze --no-pub`
- `dart analyze` in `packages/youtube_music_explode_dart`
- `dart test test/search_test.dart` in `packages/youtube_explode_dart`
- Manual checks with Michael Jackson, Arctic Monkeys, Beach Weather, Trueblood, Dominic Fike, and search/player entry paths


lib/widgets/playlist_bar.dart
```diff
@@ -77,6 +78,11 @@ class PlaylistBar extends StatelessWidget {
   bool get isFolder =>
       playlistData != null && PlaylistUtils.isFolder(playlistData!);
 
+  bool get isArtist {
+    final data = playlistData;
+    return data != null && isArtistPlaylist(data);
+  }
+
```
gokadzev, line 85: maybe we can change those getters to variables, as they don't change so no need to recall functions

lib/widgets/now_playing/now_playing_controls.dart
```diff
@@ -93,11 +98,19 @@ class NowPlayingControls extends StatelessWidget {
...
                   SizedBox(height: spacing),
                   if (metadata.artist != null)
-                    MarqueeTextWidget(
-                      text: metadata.artist!,
-                      fontColor: colorScheme.onSurfaceVariant,
-                      fontSize: artistFontSize * fontScale,
-                      fontWeight: FontWeight.w500,
+                    GestureDetector(
+                      behavior: HitTestBehavior.opaque,
+                      onTap: canOpenArtist
```
gokadzev, line 103: onTap: canOpenArtist
      ? () => _openArtistPage(context, metadata)
      : null,

lib/widgets/now_playing/now_playing_controls.dart
```diff
@@ -121,6 +134,64 @@ class NowPlayingControls extends StatelessWidget {
...
+    if (lookup.isEmpty) return;
+
+    final router = GoRouter.of(context);
+    final basePath = _artistRouteBasePath(context);
+    final artistData = {
+      'ytid': artistId.isNotEmpty ? artistId : lookup,
+      if (artist.isNotEmpty) 'title': artist,
+      if (sourceSongId.isNotEmpty) 'sourceSongId': sourceSongId,
+      if (videoAuthor.isNotEmpty) 'videoAuthor': videoAuthor,
+      'source': 'youtube-artist',
```
gokadzev, line 166: is the new source considered by other pages too? If I recall correctly, we have some conditional checks in other pages and functions too that runs based on what's the source

lib/services/playlist_download_service.dart
```diff
@@ -154,6 +154,7 @@ class OfflinePlaylistService {
           progressNotifier.value.completed > progressNotifier.value.failed) {
         // Create an offline version of the playlist
         final offlinePlaylist = {
+          ...playlist,
```
gokadzev, line 157: what's this for?

lib/screens/library_page.dart
```diff
@@ -364,6 +380,7 @@ class _LibraryPageState extends State<LibraryPage> {
     bool isOfflinePlaylists = false,
     bool hasItemsAfter = false,
     bool hasItemsBefore = false,
+    bool isArtistList = false,
```
gokadzev: we could add a new playlist source type so we don't have to pass it in the function and it carries the type everywhere.

lib/screens/library_page.dart
```diff
@@ -445,12 +458,12 @@ class _LibraryPageState extends State<LibraryPage> {
       padding: hasItemsAfter ? EdgeInsets.zero : commonListViewBottomPadding,
       itemBuilder: (BuildContext context, index) {
         final playlist = playlists[index];
-        final borderRadius = getItemBorderRadius(
-          index,
-          playlists.length,
-          hasItemsBefore: hasItemsBefore,
-          hasItemsAfter: hasItemsAfter,
-        );
+        final isLastItem = index == playlists.length - 1;
```
gokadzev: why are we coming up custom checks again? did getItemBorderRadius not handle it properly?

lib/widgets/playlist_bar.dart
elias001011: Good point. I can switch these to local finals if we want, but they only depend on playlistData and are cheap getters, so I kept them as accessors to keep build a bit cleaner.

lib/widgets/now_playing/now_playing_controls.dart
elias001011: Yes, that tap is intentional. It makes the artist label open the artist page only when we have enough metadata and are not offline.

lib/widgets/now_playing/now_playing_controls.dart
elias001011: Yes, the new source is already covered. PlaylistUtils.isArtistPlaylist treats youtube-artist as an artist source, and the artist page sets both source and isArtist, so downstream checks still see it as artist content.

lib/services/playlist_download_service.dart
elias001011: I am spreading the original playlist map so the offline copy keeps the existing metadata and then I override only the fields that need to change for the downloaded version.

lib/screens/library_page.dart
elias001011: Good point. I used a bool here to keep the initial change small because only the artist section needs it right now, but I agree an enum or source type would scale better if we add more list variants later.

lib/screens/library_page.dart
elias001011: The helper only handles first/middle/last in a flat list. This list lives inside sectioned slivers, so I needed the extra boundary checks to keep the card corners correct. If we want to simplify it, I would extend the helper to accept section-boundary flags.

lib/screens/library_page.dart
elias001011: Implemented. I replaced the raw bool with a private LibraryPage list kind enum, so the artist section is typed explicitly without threading a loose flag around.

lib/screens/library_page.dart
elias001011: Implemented. I removed the custom border math and now use getItemBorderRadius with the section boundary flags, which keeps the card corners correct without the ad-hoc checks.

lib/widgets/now_playing/now_playing_controls.dart
gokadzev: > Yes, that tap is intentional. It makes the artist label open the artist page only when we have enough metadata and are not offline.

yeah, I suggested using an arrow function instead of block function, as we don't have anything else there.

lib/screens/library_page.dart
gokadzev: > Implemented. I replaced the raw bool with a private LibraryPage list kind enum, so the artist section is typed explicitly without threading a loose flag around.

Actually i meant using 'source': 'youtube-artist' check instead, as playlists have it anyway.

elias001011: Done. I removed the extra list-kind enum and now derive artist rows from the playlist source (`youtube-artist`) directly in the builder, so the section still behaves the same without carrying extra state through the helper.

lib/widgets/now_playing/now_playing_controls.dart
elias001011: Done. I switched the tap callback to the arrow form.

PR #873, opened 2026-06-18
Keep current song when clearing queue

elias001011:
What changed
- Clear queue now preserves the currently playing track instead of emptying the queue entirely.
- Playback state is refreshed after the clear so the queue index stays aligned with the actual track that is still playing.

Why
- The old behavior removed every item and reset the queue index, which made the next song added look like the active track even though the previous song was still playing.

Impact
- After Clear queue, the player keeps showing the real current song.
- New queue additions append after that track instead of stealing the active slot.

Checks
- `flutter analyze lib/services/audio_service.dart`
- `flutter test test/widget_test.dart` (fails in the existing app smoke test because Hive boxes are not opened there)

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -1052,13 +1052,25 @@ class MusifyAudioHandler extends BaseAudioHandler {
 
   void clearQueue() {
     try {
+      final currentSong = _currentQueueIndex >= 0 &&
+              _currentQueueIndex < _queueList.length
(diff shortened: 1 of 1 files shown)


elias001011:
https://github.com/user-attachments/assets/301c00da-5488-4879-91a3-7b174e19f6b3

This is how it works today. Clearing the queue makes it disappear, and if you add a song to it, the song that's currently playing will count as if it's playing a different song when you add  in queue.

The PR prevents the current sound from being removed, thus fixing the bug.

PR #874, opened 2026-06-18
feat: add end-of-song option to sleep timer

difagume:
### Summary

Adds an "End of song" option to the sleep timer. Instead of stopping after a fixed duration, the timer now stops playback when the current song finishes.

### What changed

- New `sleepTimerEndOfSong` flag and `setSleepTimerEndOfSong()` method in `MusifyAudioHandler`
- Detects song completion in `_handleProcessingStateChange` and stops playback when the flag is active
- New "End of song" chip in the sleep timer dialog
- Regular sleep timer now uses `stop()` instead of `pause()` for consistent behavior across both modes
- Fixed: sleep timer button no longer stays highlighted after expiration (`Duration.zero` → `null`)
- Fixed: `sleepTimerExpired` is reset on `ProcessingState.ready`, so auto-advance works if the user resumes playback after the timer fired
- Localized in 21 languages

### Screenshots
Screenshot

### DIFF
--- a/lib/localization/app_de.arb
+++ b/lib/localization/app_de.arb
@@ -214,6 +214,7 @@
   "sleepTimer": "Sleep timer",
   "sleepTimerCancelled": "Sleep-Timer abgebrochen",
   "sleepTimerSet": "Sleep-Timer",
+  "endOfSong": "Ende des Songs",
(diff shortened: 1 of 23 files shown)


gokadzev:
@difagume Thanks!

PR #892, opened 2026-07-11
enhancement: cache SponsorBlock segments for offline playback

elias001011:
When a song is downloaded for offline playback, SponsorBlock segments are lost — the app stops making HTTP requests offline, so skip segments are never applied.

This PR stores the SponsorBlock segments of a song at download time and reads them back during offline playback, so offline songs still respect SponsorBlock skips. Songs downloaded before this existed are looked up lazily on their first offline play (when online).

### Where the segments live

They are stored inside the song's own record in the `userNoBackup` box, right next to `audioPath`, so they last exactly as long as the download does: they are not backed up, they never expire, and removing the song removes them. There is no separate cache to keep in sync.

### Changes

- `common_services.dart`:
  - `cacheSponsorBlockSegments()` — looks a song up once and stores the result during `makeSongOffline()`. A confirmed empty result is stored too, so a song known to have no segments is never looked up again.
  - `getCachedSponsorBlockSegments()` — synchronous, offline-safe read. `null` means nothing stored yet; `[]` means looked up and confirmed to have none.
  - `_fetchSkipSegments()` — the request itself, split out of `getSkipSegments()`. It throws on failure instead of swallowing it, which is what lets the offline cache tell "confirmed no segments" apart from "failed to fetch". `getSkipSegments()` keeps its previous behaviour (returns `[]` on any error) for the online path.
  - `_sponsorSerialCall()` — serializes all SponsorBlock API calls through a single `Future` chain, so one request is in flight at a time, spaced 1s apart.
- `audio_service.dart`:
  - `_buildSkippedAudioSource()` — the clipping logic, extracted so the online and offline paths share it instead of duplicating it.
  - `_applyOfflineSponsorBlock()` — applies stored segments to an offline audio source, and kicks off a background lookup when a song has nothing stored yet.

### Not storing false negatives

The distinction between "this video genuinely has no segments" and "the lookup failed" matters, because a stored empty result is permanent — it is never looked up again. So:

- **404 / `Not Found`** is how the API answers "no segments for this video". That is a real empty result and gets stored.
- **Any other non-200, a network error, or a parse error** throws. `cacheSponsorBlockSegments()` then retries once (spaced by the queue below), and if that fails too it stores nothing, so the song is simply looked up again the next time it plays.

### Rate limiting

SponsorBlock's public API (`sponsor.ajay.app`) documents no official rate limit, but it is a shared unauthenticated third-party endpoint, so this is conservative about how often it is hit:

- **One lookup per song, ever** — including songs confirmed to have no segments.
- **Serialized requests** — every lookup goes through a single `Future` chain (`_sponsorSerialCall`) that guarantees one request in flight at a time, 1s apart. This matters because `cacheSponsorBlockSegments()` is called with `unawaited(...)` from `makeSongOffline()`, so it does not block the download queue — several songs can finish downloading close together (a large playlist download runs 3 concurrent workers) and each fires its own lookup independently. An earlier revision used a timestamp throttle instead of a real queue, which had a race: concurrent calls could all read the same "last call" timestamp before any of them updated it and end up firing together. The `Future`-chain approach fixes that by construction, since Dart's event loop guarantees no interleaving between reading and reassigning the chain.

This makes the feature safe on very large playlists (500+ songs) without a burst of simultaneous requests to a third-party API.

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -2248,7 +2248,14 @@ class MusifyAudioHandler extends BaseAudioHandler {
       final tag = mapToMediaItem(song);
 
       if (isOffline) {
-        return AudioSource.file(songUrl, tag: tag);
+        final fileSource = AudioSource.file(songUrl, tag: tag);
+
+        if (sponsorBlockSupport.value) {
+          return _applyOfflineSponsorBlock(fileSource, song['ytid']) ??
+              fileSource;
+        }
+
+        return fileSource;
       }
 
       final uri = Uri.parse(songUrl);
@@ -2273,6 +2280,22 @@ class MusifyAudioHandler extends BaseAudioHandler {
     }
   }
 
+  AudioSource? _applyOfflineSponsorBlock(
+    UriAudioSource audioSource,
+    String songId,
+  ) {
+    final segments = getCachedSponsorBlockSegments(songId);
+    if (segments != null) {
+      return segments.isEmpty
+          ? null
+          : _buildSkippedAudioSource(audioSource, segments);
+    }
+    // Nothing stored yet, e.g. a song downloaded before this existed: look it
+    // up now so the next playback of it skips offline too.
+    if (!offlineMode.value) unawaited(cacheSponsorBlockSegments(songId));
+    return null;
+  }
+
   Future<AudioSource?> checkIfSponsorBlockIsAvailable(
     UriAudioSource audioSource,
     String songId,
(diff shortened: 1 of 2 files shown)


gokadzev:
@elias001011 Hi, I think the code is too much overcomplicated + we already have database functions to get/add/update caches

elias001011:
Hi @gokadzev, thanks for the feedback! You're right — `cacheSponsorBlockSegments()` and `getCachedSponsorBlockSegments()` were reaching into `Hive.box('cache')` directly instead of using the existing `getData`/`addOrUpdateData` helpers in `data_manager.dart`. Fixed in e4d407e: both now go through those helpers, which also gives the cache TTL/expiry handling for free instead of caching segments forever.

Left the serial-call queue (`_sponsorSerialCall`) and retry-once logic as-is, since that part isn't touching the db layer — it just spaces out SponsorBlock API calls when several songs finish downloading close together (e.g. a big playlist download with 3 concurrent workers). Happy to simplify that further too if you think it's still more than needed.

PR #895, opened 2026-07-13
Feat/sleep timer fadeout

elias001011:
Sleep timer now fades playback out instead of stopping abruptly.

### Changes

- `setSleepTimer()`: the timer fires 6 seconds before the target duration, then hands off to a fade instead of calling `stop()` directly — so the total time-to-silence still matches what the user set.
- `_performSleepTimerFadeOut()`: linearly ramps volume down to 0 over 6 seconds (20 steps), then stops playback and restores volume. The fade is anchored to elapsed wall-clock time rather than summing fixed per-step delays, so the platform-channel overhead of each `setVolume` call doesn't accumulate into extra seconds by the end of the fade.
- Captures the volume in place before the fade starts and restores that value afterward, instead of assuming full volume.

### Behavior around interruptions

The timer (and its fade) are intentionally decoupled from playback controls — pausing, skipping, adding to the queue, or changing repeat mode never cancels it. This matches the app's existing pre-fade sleep timer behavior, and how Spotify's own sleep timer works: it keeps counting down and stops on schedule no matter what's currently playing.

The three actions that *do* take over a running fade all stop it and restore the volume immediately, rather than waiting for its next step to notice:

- **Cancelling** the sleep timer — clears everything, as before.
- **Setting a new timer** during the fade window — the new timer takes over. A fade that has been replaced no longer clears the timer state on its way out, which would otherwise leave the button reading as inactive while playback was still scheduled to stop.
- **Switching to end-of-song** during the fade window — the fade stops instead of continuing to ramp down and calling `stop()` before the song actually ends.

If the fade itself throws partway through, the volume is put back, so a failed fade can't leave playback muted with nothing in the UI able to undo it.

Tested on-device over wireless debugging: timer fires, fade is audible and click-free, and skipping/pausing/queueing during the fade window no longer cancels the timer.

### DIFF
--- a/lib/services/audio_service.dart
+++ b/lib/services/audio_service.dart
@@ -74,6 +74,11 @@ class MusifyAudioHandler extends BaseAudioHandler {
   Timer? _debounceTimer;
   bool sleepTimerExpired = false;
   bool sleepTimerEndOfSong = false;
+  bool _sleepTimerFading = false;
+  double _preFadeVolume = 1;
+
+  static const int _sleepTimerFadeSteps = 20;
+  static const Duration _sleepTimerFadeDuration = Duration(seconds: 6);
 
   final List<Map> _queueList = [];
   final List<Map> _originalQueueList = [];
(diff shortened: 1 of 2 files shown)


elias001011:
Closing this — it's fallen ~87 commits behind and both touched files (`audio_service.dart`, `bottom_actions_row.dart`) have since been reworked upstream, so it no longer applies cleanly. The implementation is also heavier than it needs to be (volume save/restore is dead weight since the app never changes volume; the manual wall-clock-anchored fade loop can be a plain `Timer.periodic`). If the fade-out is still wanted I'll re-cut a minimal version against current master.

PR #897, opened 2026-07-15
Translation hu

peterczegledy:


gokadzev:
@peterczegledy Hi, you can use Crowdin to translate files: https://crowdin.com/project/musify

PR #904, opened 2026-07-29
feat: artist landing page with top songs and discography

elias001011:
> **Status:** rebased on master, both suggestions from your review applied, and
> a review pass of my own on top. Ready to merge from my side, with the spacing
> between sections left to you as agreed. Two of the commits here are yours: the
> `song_bar` artwork fix and the `resolvePlaylist` catch — thanks, both look
> right to me. What changed since you last read this is at the bottom.

## What this does

Opening an artist currently waits for the whole catalog to be walked album by
album, track by track, before anything is shown. This adds a landing page built
from one YouTube Music browse request plus its release grids:

- header with avatar, monthly listeners and bio
- top songs, ranked, with the YouTube Music play count
- the discography, split in albums and singles & EPs
- suggested artists — these were already in the same browse response, the parser
  was walking those nodes and discarding them
- releases open as playlists
- **"All songs" still opens the existing full artist playlist, unchanged**

The existing catalog infrastructure is untouched. This is a layer on top of it,
not a replacement.

## How it fits with the existing artist playlist

The rule, in one sentence: **the landing page is the online layer; downloading
an artist downloads its songs, not the page.**

| Situation | What happens |
|---|---|
| Online | Landing page |
| Play / shuffle pressed | Reads the "All songs" catalog, then plays it |
| Download pressed | Same catalog, stored under the artist id |
| Offline, downloaded | Landing page is never shown — the artist opens as its downloaded songs |
| Offline, not downloaded | Redirected home, as before |

So there is no "offline artist page" state to keep in sync. `artist_profile` and
`artist_album` are network caches only.

Refreshing either page refreshes the artist page and marks the song catalog
stale, so the two cannot disagree: refreshing from "All songs" rebuilds the
catalog there and then, refreshing the landing page rebuilds it the next time
the songs are read. The exception is a **downloaded** artist, whose "All songs"
answers from the download until it is refreshed from that page — unchanged from
before this branch, and the reason the landing page shows a new release before
"All songs" does.

Both pages read the songs through the same `getPlaylistInfoForWidget` call, so
the landing page and "All songs" always agree — including the case where the
artist was downloaded, where both use the downloaded copy.

## What it costs

Measured against the live API on an artist with 122 releases (Taylor Swift):

| | Requests | Time |
|---|---|---|
| Landing page | 1 + 2 grids, in parallel | ~3.4 s cold, cached afterwards |
| A release | 1 | cached per release |
| "All songs" / play / shuffle | 1 per release, 6 at a time | ~14–30 s cold, ~40 ms cached |

The catalog walk is the pre-existing cost of the artist playlist, not something
this PR adds — it is just no longer paid before the page can render. Album
responses are now cached per release, so opening a few releases warms part of
it, and pressing play once makes "All songs" instant.

## Reuse

Trying to keep this from being a pile of new surface:

- the playlist **download button**, **like button** and **play/shuffle row**
  moved out of `playlist_page` and are shared with this page — that file is
  **229 lines shorter**. The play/shuffle row is `PlaylistActionButtons`, which
  already existed in the tree unused. The moved code is unchanged for playlists;
  what an artist does differently lives at the call site (the download button is
  given a resolver instead of an on-screen list, because the landing page does
  not hold one). One deliberate difference: the like button now says *liked
  playlists* instead of *liked songs*, using strings that already existed —
  `playlist_page` was passing the song wording to a button that likes a playlist
- the artist header **is** `PlaylistHeader`, which already had an `isArtist`
  branch; it grew an optional monthly-listeners chip and bio
- a release **is** a playlist: `getPlaylistInfoForWidget` recognises `MPRE` ids
  and `getAlbum` reads the release's own title, artwork and artist, so there is
  no album page and no metadata to thread through the router. Deep links and
  liked releases work as a result
- `getArtistReleases` and `getArtistTopTracks` fold into `getArtistProfile`:
  one call answers everything about an artist, so the landing page and the
  catalog cannot disagree about the discography
- `_routeBasePath` was copied in three files, now
  `NavigationManager.artistPath` / `artistSongsPath` / `albumPath`
- both pages use the existing `AsyncLoader` instead of hand-rolled
  `FutureBuilder` + spinner + error

## Fixed since the first push

- **Play and shuffle now play the artist.** The "Top songs" shelf holds exactly
  five rows, which is what playback was getting. The progressive
  queue-growing mechanism is gone: the buttons read the same catalog "All songs"
  holds, showing a spinner while they do. That also removes the dozens of
  background requests that were queueing songs while the page was being used.
- **Suggested artists are verified** against a real response ("Fans might also
  like", 10 entries, matched by `UC` browse id + artist page type).
- Release types and years are verified too: every release of the test artist
  comes back typed (31 albums / 72 singles / 19 EPs) and dated, so the shelf
  fallback heuristics the first version needed are gone.
- Offline no longer fires a network lookup when the page is built.
- The play count is stored next to the song instead of inside it, so it cannot
  ride along into the queue or the library with the song map.
- The `catalogStatus: 'loading'` flag the two pages passed between each other is
  gone; a failed catalog no longer leaves a permanent spinner.

## One call I would like yours on

The download button had to be shared for this page to exist at all, so I did the
same with the like button: `playlist_page` ends up 45 lines shorter and the two
pages cannot drift apart on a liked artist, which is the one piece of state they
both write. The cost is that it touches a button every playlist and album page
uses, and its tooltip now reads *liked playlists* instead of *liked songs*.

If you would rather keep `playlist_page` untouched, dropping the like extraction
is self-contained — the landing page keeps its own copy and nothing else in the
PR depends on it. The same reasoning is why the **sync** button is deliberately
*not* shared: the two pages refresh different things (the landing refreshes the
artist page and marks the song catalog stale, "All songs" rebuilds that catalog),
so only the icon would be common.

## Both known issues are fixed, following your review

- **Refreshing a release works.** Your conditional, plus the offline copy no
  longer short-circuiting a forced refresh — without that, Update kept doing
  nothing on a release that had been downloaded. The artist and the release
  refresh also report themselves now; they were silent, which is part of why
  the button looked dead.
- **Opening an artist from a song no longer searches first**, but not through
  `isVerifiedArtist`: that flag cannot be set from `artistId`, because the
  field holds `song.channelId`, i.e. the channel that uploaded the video, not
  the YouTube Music artist. Measured against the API: browsing Taylor Swift's
  main channel instead of her artist channel returns her page **without the
  "Top songs" shelf** (albums, singles and the related artists are there, the
  shelf is not), and browsing a label channel — NoCopyrightSounds — returns a
  page with no releases and no top songs at all. Trusting the id blindly would
  cache half a page for half the discographies, and open a page for a channel
  that is not an artist for the rest.

  What it does instead: a channel id is browsed directly, skipping the search.
  The header of that answer names the canonical artist channel through its
  subscribe button, even when the browse asked for an uploader, so the real
  page is read when the two differ and the detour is remembered. A page with
  neither releases nor top songs is not an artist, and only that case falls
  back to the search by name. A song whose channel is an artist opens it in one
  request, and in none once the page is cached; a song uploaded by a label
  still costs the search, every time, because that channel carries every artist
  the label publishes and nothing about it may be remembered.

## Unrelated bug

Fixed by your commit on this branch: both `song_bar` call sites pass `size`.

## Also in this push

- **A release is cached as itself.** The cache key did not include the artist
  its tracks are credited to, and the two readers disagree — walking a
  discography credits the songs to the artist being walked, the release page
  does not. Whichever ran first won for a week, so opening a compilation after
  pressing play on an artist showed every track as that artist's, and those
  maps travel into the queue and into the library. The crediting now happens
  after the read, so one entry answers both.
- **No release is fetched in offline mode.** A liked release opens as
  `/playlist/<MPRE…>`, which the `/album/` offline redirect does not cover.
- **The download button shows a spinner while it reads the songs.** On the
  artist page that is the whole discography and nothing on the button said so,
  so a second tap read all of it again.
- **A refresh that fails keeps the page it was refreshing.** The cached page and
  the song catalog were dropped before the request went out, so a refresh on a
  bad connection left the artist with nothing: the page became the error page,
  and the next visit had to fetch everything again. The catalog is now dropped
  once the new page is in hand, which is also the only moment it is known to be
  stale.
- Reloading the page when the route rebuilds no longer fetches in offline mode,
  the way the first load already did. It no longer reloads at all when a page is
  pushed on top of it: it was comparing the map it was navigated with, which the
  router rebuilds on every navigation, so the loader flashed over the page.
- **A refresh that fails puts the page back.** The future the page is built
  from had already been swapped for the one that failed, so a bad connection
  replaced an artist that had loaded fine with the not-found page. It reports
  the failure with a toast instead; the refresh itself is unchanged.
- **The page follows offline mode while it is open.** The setting decides which
  of the two pages this is on every build but was read once, so turning it off
  with an artist open stranded that page.
- An artist that is not found says so, instead of reporting an error of the app.
- Dead payload dropped: `sourceTitle` and `lookupTitle` were written into every
  artist map, and their only reader left with `_buildArtistCatalogFromMusic`.

Spacing between the sections is untouched, as you asked.


lib/screens/playlist_page.dart
gokadzev, line 529: Referring to the known issue N1: have you tried putting conditional check here?

like instead of just await updatePlaylistList(context, playlistId); in the line 523, we can add:

```dart
: playlistId.startsWith('MPRE') ? await getPlaylistInfoForWidget(playlistId, forceRefresh: true,) : await updatePlaylistList(context, playlistId);
```

lib/widgets/now_playing/now_playing_controls.dart
gokadzev, line 163: Referring to the known issue N2:

maybe we can try adding:

'isVerifiedArtist': info.artistId.isNotEmpty,


so resolveArtist() will skip the canonical channel search and return early

PR #905, opened 2026-07-29
Add optional audio quality badge on now playing artwork

### LINKED ISSUE #828: [Feature] Audio Quality Information Display
### Is there an existing issue for this?

- [x] I have searched the existing issues and confirmed that this feature is not already requested.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this feature.

### Category of the Feature

_No response_

### Credits

me

### Pain

When streaming music, users have no way to know the quality of the audio they are listening to. While Musify streams from YouTube which provides various audio quality levels, there's no visual feedback about what quality the user is currently experiencing. This is particularly frustrating for users who care about audio quality and want to know if they're listening to the best available version of a track.

Without this information, users cannot:
- Verify if they're getting high-quality audio (especially important for audiophiles)
- Diagnose why certain tracks might sound better or worse than others
- Make informed decisions about streaming quality settings
- Appreciate the technical aspects of their listening experience

This creates a black box experience where users blindly consume audio without any metadata about what they're actually hearing.

### Feature Overview

Display real-time audio quality information of the currently playing track in a subtle, non-intrusive location within the player interface.

### Implementation Details

1. **Information to Display**
   - Audio bitrate in kbps (e.g., "128 kbps", "256 kbps")
   - Sample rate if available (e.g., "44.1 kHz")
   - Audio codec if detectable (e.g., "AAC", "MP3", "Opus")
   - Channel configuration (Stereo/Mono)

2. **Suggested UI Locations** (choose the most appropriate):

   **Option A: Below Album Art**
   - Place quality info as subtle text below the album art
   - Use muted colors so it doesn't distract from the artwork
   - Example: "128 kbps • AAC • Stereo" in small, light gray text

   **Option B: Corner Badge on Album Art**
   - Small badge in the bottom-right or top-right corner of album art
   - Shows only the bitrate (e.g., "128k") with subtle background
   - Tap to expand and show full details

   **Option C: Info Button**
   - Small "i" icon button near the song title
   - Tapping reveals a bottom sheet or tooltip with full audio details
   - Keeps the UI clean while providing access to information

   **Option D: Swipe-up Player**
   - When swiping to expand the full-screen player, show quality info
   - Place it below the progress bar or near the time indicators
   - Example: "Audio: 128 kbps AAC Stereo @ 44.1kHz"

3. **Technical Implementation**
   - Extract audio metadata during streaming from YouTube
   - YouTube typically provides: audio bitrate (128-256 kbps), format (AAC/Opus), sample rate (44.1/48 kHz)
   - Display "Unknown" if metadata cannot be retrieved
   - Update info when track changes or quality switches

4. **Visual Design**
   - Minimalist style that blends with existing UI
   - Use system font at appropriate size (12-14sp for secondary info)
   - Muted color that doesn't compete with primary UI elements
   - Consider dark/light theme variants

5. **User Experience Considerations**
   - Not intrusive - should not distract from listening
   - Optional - user can disable in settings if they prefer clean UI
   - Tooltip/help text explaining what the values mean on first tap
   - Long-press to see detailed explanation of audio quality concepts

### Useful Resources/References

- [YouTube Media Delivery - Audio Quality](https://developers.google.com/youtube/v3/live-streaming-errors): Information about YouTube audio streams and quality levels.
- [Audio Bitrate Guide](https://www.diffen.com/difference/Audio_Bitrate): Explanation of audio bitrates and their impact on quality.
- Similar implementations: Spotify shows "Very High" quality indicator, Tidal displays detailed audio information including lossless formats.

elias001011:
## Summary
- Adds an off-by-default badge on the now playing artwork showing the current song's bitrate/codec (e.g. `128 kbps • AAC`), toggled right next to the audio quality picker in Settings.
- **Works both online and offline.** For a downloaded song, the badge reads the bitrate/codec captured once at download time (stored alongside `audioPath` in the offline song record) — zero network calls, works with no connection. For a song streaming online, it reuses the stream already resolved for playback, skipped entirely when the global Offline mode setting is on.
- **Costs no extra request.** Following review feedback, the audio stream selected for a song is now kept in an in-memory map in `common_services`, keyed by song id and audio quality setting, rather than only its URL reaching the Hive cache. `fetchSongStreamUrl` goes through `fetchBestAudioStream`, so the stream manifest is fetched once per song no matter how many callers need it, and the badge adds nothing on top once playback has resolved the song. Entries expire on the same window the URL cache already treated as fresh, and the map is capped so it can't grow unbounded over a long session.
- Kept minimal otherwise: no listener on playback state, and the only touch to the playback path is the retry branch now dropping both caches together (`invalidateSongStreamCache`) instead of just the URL key — without that a retry would reuse the stale stream and hit the same dead URL.

## Known limitation
Songs downloaded **before** this change won't have `audioBitrateKbps`/`audioCodec` stored in their offline record, so the badge won't show for them until they're re-downloaded. Intentionally not adding a backfill/migration step for this — fetching quality info in bulk for the whole existing download library felt like unnecessary weight for a purely cosmetic badge. It self-resolves naturally as songs get re-downloaded.

Closes #828

### DIFF
--- a/lib/widgets/now_playing/now_playing_artwork.dart
+++ b/lib/widgets/now_playing/now_playing_artwork.dart
@@ -19,6 +19,8 @@
  *     please visit: https://github.com/gokadzev/Musify
  */
 
+import 'dart:async';
+
 import 'package:audio_service/audio_service.dart';
 import 'package:fluentui_system_icons/fluentui_system_icons.dart';
 import 'package:flutter/material.dart';
@@ -75,14 +77,23 @@ class NowPlayingArtwork extends StatelessWidget {
             ),
           ],
         ),
-        child: ClipRRect(
-          borderRadius: BorderRadius.circular(borderRadius),
-          child: SongArtworkWidget(
-            metadata: metadata,
-            size: imageSize,
-            errorWidgetIconSize: size.width / 8,
-            borderRadius: borderRadius,
-          ),
+        child: Stack(
+          children: [
+            ClipRRect(
+              borderRadius: BorderRadius.circular(borderRadius),
+              child: SongArtworkWidget(
+                metadata: metadata,
+                size: imageSize,
+                errorWidgetIconSize: size.width / 8,
+                borderRadius: borderRadius,
+              ),
+            ),
+            Positioned(
+              right: 10,
+              bottom: 10,
+              child: _AudioQualityBadge(metadata: metadata),
+            ),
+          ],
         ),
       ),
       backWidget: Container(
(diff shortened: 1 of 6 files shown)


lib/widgets/now_playing/now_playing_artwork.dart
```diff
@@ -169,3 +181,122 @@ class NowPlayingArtwork extends StatelessWidget {
     );
   }
 }
+
+typedef _AudioQualityInfo = ({int bitrateKbps, String codec});
+
+String _normalizeCodec(String codec) => switch (codec) {
```
gokadzev, line 186: Are we sure we need this?

lib/widgets/now_playing/now_playing_artwork.dart
```diff
@@ -169,3 +181,122 @@ class NowPlayingArtwork extends StatelessWidget {
...
+  final c when c.startsWith('opus') => 'Opus',
+  final c when c.startsWith('mp4a') => 'AAC',
+  final c => c,
+};
+
+/// Resolves once per song: reads the quality stored at download time for a
```
gokadzev, line 192: maybe we can use different caching for stream url (for instance map) so we don't need to fetch a stream twice?

elias001011: Good call — done, and it ended up removing code rather than adding it.

The selected `AudioOnlyStreamInfo` now lives in an in-memory map in `common_services`, keyed by song id + quality setting. `fetchBestAudioStream` checks it before touching the manifest, and `fetchSongStreamUrl` no longer fetches the manifest itself — it goes through `fetchBestAudioStream` and just persists the URL to Hive as before. So the manifest is fetched once per song no matter how many callers need it, and the badge costs no request at all once playback has resolved the song.

The separate Hive `audioQuality_*` cache I had in the widget is gone.

Two details worth flagging:
- Entries expire after the same window the URL cache already treated as fresh (`_cacheValidationDuration`), and the map is capped at 50 entries so it can't grow unbounded over a long session.
- The retry path in `audio_service` that used to `deleteData` the URL key now calls `invalidateSongStreamCache`, which drops both. Without that, a retry after a dead URL would reuse the stale stream from the map and hit the same broken URL again.

lib/widgets/now_playing/now_playing_artwork.dart
elias001011: The `typedef` is just an alias so the record type isn't spelled out in three places — no behaviour attached to it, happy to drop it and write the type inline if you prefer.

`_normalizeCodec` does pull its weight though: the manifest hands back raw codec strings like `mp4a.40.2` and `opus`, so without it the badge literally reads `128 kbps • mp4a.40.2`. It maps those two to `AAC`/`Opus` and passes anything unrecognised through unchanged. If you'd rather show the raw value and skip the mapping, say the word and I'll remove it.

PR #943, opened 2026-09-02
fix: match radio station card background to the rest of the app

elias001011:
## Problem

On the radio stations page (and radio results in search), `RadioStationCard` renders on `colorScheme.surfaceContainer`, one tonal step above every other list row in the app. `SongBar`, `PlaylistBar` and `CustomBar` all use `colorScheme.surfaceContainerLow`, so the radio cards look like a heavier, more opaque surface that doesn't blend with the rest of the UI.

## Fix

Use `colorScheme.surfaceContainerLow` for `RadioStationCard`, matching the shared list-row background.

One line. `flutter analyze` clean.

### DIFF
--- a/lib/widgets/radio_station_card.dart
+++ b/lib/widgets/radio_station_card.dart
@@ -90,7 +90,7 @@ class _RadioStationCardState extends State<RadioStationCard> {
 
     return Card(
       elevation: 0,
-      color: colorScheme.surfaceContainer,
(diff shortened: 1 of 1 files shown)


elias001011:
Closing — this changed the wrong thing. The actual bug is the mini player leaving a blank strip on the radio stations page (the page reserves mini-player space via Scaffold.bottomNavigationBar instead of inside the scroll content). Replacement PR incoming.

PR #947, opened 2026-09-07
Add the Android Auto browse and search interface

vysakh-i:
The manifest already registered a MediaBrowserService and declared the app to Android Auto, and the handler already answered getChildren, so the car could connect and draw a browse tree. Tapping anything in it did nothing.

Browse items were built by mapToMediaItem, whose id comes from song['id'] - a list index assigned by returnSongLayout, not a stable identifier. Playback resolved that id with _findSongByYtid, which looks for a song whose ytid is "3". It never matched, so every tap was a silent no-op. Queue rows were worse: their ids are queue entry ids, which no ytid lookup can match either.

Media ids now carry the container a song was listed under plus its token within that container, so a tap can resolve back to the exact row and enqueue the rest of the list behind it. Skip-next from the wheel now walks the list the driver was looking at instead of dead-ending on a single track. The queue is keyed by queue entry id rather than ytid, so tapping the second copy of a song does not jump to the first.

Also fills in the rest of the interface that was missing:

- search() was never implemented, so the car's search screen was always empty. It now searches the device first, which answers instantly and works with no signal, then tops up from YouTube.
- playFromSearch only matched already-local songs, so most voice requests failed. It reuses the same search and queues every hit, not just the best guess.
- subscribeToChildren was never implemented, so the browse tree was frozen at whatever it held when the car connected. Each root now republishes when its backing data changes. Queue notifications are throttled, since the queue emits several times per song.
- Playlists were not browsable at all; user-created ones are listed first as they are local. A user-created playlist is inline or empty, so it never triggers an online fetch against its local timestamp id.
- androidBrowsableRootExtras was unset, so Auto was told neither that content style is supported nor that search exists, and never offered a search box.
- Empty categories rendered as a blank screen, which reads as a broken app, and now say why they are empty.

Verified with flutter analyze (clean, whole project) and dart format on Flutter 3.47.2 / Dart 3.13.2, matching the pubspec constraint.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -292,6 +292,17 @@ Future<void> initialisation() async {
         androidNotificationIcon: 'drawable/ic_launcher_foreground',
         androidShowNotificationBadge: true,
         androidStopForegroundOnPause: false,
+        // Handed to Android Auto on the browsable root. Without the search
+        // flag the car never offers a search box, and without the content
+        // style keys it falls back to an undifferentiated list of everything.
+        androidBrowsableRootExtras: {
+          AndroidContentStyle.supportedKey: true,
+          AndroidContentStyle.browsableHintKey:
+              AndroidContentStyle.gridItemHintValue,
+          AndroidContentStyle.playableHintKey:
+              AndroidContentStyle.listItemHintValue,
+          'android.media.browse.SEARCH_SUPPORTED': true,
+        },
       ),
     );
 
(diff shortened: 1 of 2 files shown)


gokadzev:
@vysakh-i Hi, is this tested on real devices?

vysakh-i:
> @vysakh-i Hi, is this tested on real devices?

Yes, tested in my vehicle and seems to be working well.

gokadzev:
Thanks!

PR #954, opened 2026-09-17
fix(stats): stop counting listening time after the queue ends

### LINKED ISSUE #953: [Bug] Song minutes in Time Machine still count up when the end of a song is reached
### Acknowledgements

- [x] I have searched the existing issues and confirmed that this is a new issue, not a duplicate or related to another open issue.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to the latest version. You can find the latest version [here](https://github.com/gokadzev/Musify/releases).
- [x] I will fill out all the sections in this issue report.

### Current Behavior

When a song reaches its end but no other song plays (such as at the end of the queue), the song will stop, but the "Minutes listened" stat in Time Machine continues to increase.

### Expected Behavior

When a song stops playing, no more minutes would be recorded as listened.

### Steps To Reproduce

1. Make a queue with only one song
2. Skip to the end of the song
3. Wait a few minutes
4. Observe change in Minutes Listened by refreshing it, such as restarting the app or pressing "Listening Stats" at the bottom.

### Musify Version

V10.4.0

### Device And Operating System

Lenovo Idea Tab

### Android Version You Use

Android 16

### Additional Information

_No response_

elias001011:
Fixes #953

### Problem
With a single-song queue (or any end of queue with repeat/auto-play off), "Minutes listened" in Time Machine keeps growing after the song ends.

### Cause
`just_audio` keeps `player.playing == true` after a track completes, until `pause()`/`stop()` is called (documented on `AudioPlayer.play`). When the last song ends:

1. `processingStateStream` → `completed` → `_handleProcessingStateChange` correctly finishes the listening session.
2. `skipToNext()` has nowhere to go, so the player stays in `(playing: true, completed)`.
3. `playerStateStream` delivers that state to `handlePlayerStateForListeningStats`, which sees a `false → true` "playing" transition and starts a **new** session for the same song.
4. Every later checkpoint (pause, seek, app backgrounding) then records wall-clock idle time as listened.

### Fix
In `ListeningStatsService.handlePlayerStateForListeningStats`, treat `completed`/`idle` as not playing, so no session is resumed/started while no audio is actually playing. Replay after completion goes through `playAgain()`, which starts a fresh session as before.

### Tests
- `test/listening_stats_end_of_queue_test.dart` drives the real `ListeningStatsService` through the exact event sequence above. Without the fix, the first test reproduces the bug (3 s counted vs 1 s real playback); with the fix all three pass.
- `flutter analyze` clean on the changed files; full `flutter test` unchanged apart from the pre-existing `widget_test.dart` boilerplate.


elias001011:
Small one, but it was a sneaky bug — thanks to @Xreturns for the clear repro steps. Happy to adjust anything if you'd like it done differently.

Issue #894, opened 2026-07-11
[Feature] Persistent Queue


sethro57:
### Is there an existing issue for this?

- [x] I have searched the existing issues and confirmed that this feature is not already requested.
- [x] I read the [FAQ section](https://github.com/gokadzev/Musify/discussions/340) and there's nothing mentioned about this feature.

### Category of the Feature

UX Improvements

### Credits

myself

### Pain

Any time I close the app, the queue for my shuffle is wiped, forcing me to create a new queue every time I re-open the app.

### Suggested Solution

Add a persistent queue. When this option is enabled, it should keep track of your current queue, as well as your current position, and restore it when you open the app again.

### Useful Resources/References

YouTube Music itself has this feature, as well as other YTM clients.

PR #956, opened 2026-09-18
feat(playback): keep the queue across app restarts (closes #894).

nicotontige:
### The problem

Closing the app throws the queue away. A shuffle put together over an evening
is gone at the next launch, and the only thing left to do is build it again.

### What this does

The queue is written to storage as it changes, along with the song being
played and how far into it playback got, and comes back at the next launch.
The songs, the shuffled order and the pre-shuffle one are all kept, so turning
shuffle off after a restart still restores the original order. Repeat and
shuffle modes were already persisted and are unaffected.

### How it works

`QueuePersistenceService` (new, `lib/services/queue_persistence_service.dart`)
owns the storage; the audio handler only calls into it.

- **Nothing is fetched while restoring.** At startup the queue is published so
  the player shows where it stopped, nothing more — no stream URL is resolved,
  no network call is made, and playback does not start on its own. The source
  is only built when playback is asked for, and resumes through
  `setAudioSource(initialPosition:)`.
- **A position belongs to one queue entry.** It is only honoured for the entry
  id it was taken on, and only once. A queue edited in between restores from
  the start of the song rather than dropping the listener into the middle of a
  different one.
- **The song list and the position are stored apart.** The list is only
  rewritten when it actually changes; the position at most every ten seconds of
  playback. A long queue is therefore not rewritten on every tick. Both are
  flushed when playback pauses or stops and when the app leaves the foreground
  — including `onTaskRemoved`, which is what a swiped-away app goes out
  through, and the flush happens before `audioPlayer.stop()` takes the position
  back to zero.
- **Bounded.** Stored in the `userNoBackup` box, so it is not carried into
  backups. A queue beyond 1000 songs is stored as a window around the song
  being played.

One behaviour change outside the queue itself: `play()` with no loaded source
now resumes the current queue instead of replacing it with a recently played
song. That path previously discarded a queue that had failed to load its
source.

### Not behind a setting

The issue asked for this to be opt-in. I built it that way first, then dropped
the toggle: bringing the queue back is what an app is expected to do when it
opens, and a queue that survives is the ordinary case rather than something to
go looking for in the settings. The toggle is a small addition if you would
rather have it — say the word and I will put it back.

### Testing

- `flutter analyze`: clean.
- `flutter test`: six new tests in `test/queue_persistence_test.dart` cover the
  round trip, the shuffle order, a position whose song moved, the save that
  follows a restore, the oversized-queue window, and an emptied queue. All
  pass. (`test/widget_test.dart`, the stock counter template, fails on master
  as well.)
- Profile APK builds and installs.


elias001011:
One thing I noticed: #894 originally described the persistent queue as an opt-in setting, but this PR removes the toggle and makes it the default behavior for everyone.

I don't necessarily think that's wrong, but it seems like a product/behavior change beyond the original issue. Was making it always-on intentional?

nicotontige:
Yes, this was intentional. Nothing prevents you from playing a new playlist as usual when opening the app, but if the user wants to, they can simply pick up where they left off in their playlist.
It's also the default behavior in many apps like Deezer, Spotify, and YouTube Music, so it made sense to make it the default here too.

